PR #1701, opened 2013-01-26
A @nosandbox Option . . . 

komichi:
These two commits add:

  (1) a @nosandbox option to user scripts which, when enabled, allows execution with full DOM _and_ xmlHttpRequest access
  (2) a per-script userNosandbox option (similar to User Includes/Excludes) that a user must check in order to allow the @nosandbox option before it is actually enabled

In other words, the first commit enables @nosandbox in user scripts and any script with @nosandbox set to true runs without the typical access wrapping (i.e., window == unsafeWindow) and has full access to all granted GM_* functions.  The second commit makes @nosandbox a per-script user option that is off by default; if both @nosandbox and the script.userNosandbox options are not true, the current ("as safe as possible") behavior holds.

Having followed some of the discussions here and on the mailing list, I understand the need for sandboxing user scripts and respect your efforts at implementing better GM security.  However, there are many cases when full DOM and AJAX request access is a requirement and I do not see a better way than the above two commits at allowing this and still pleasing everyone (please correct me if I am wrong here).

For instance, the main use case is typically something like: you fully trust the site you are running GM on (e.g., you may even have full source access) but for corporate, administrative or political reasons, you simply cannot get any changes to the site implemented.

Please consider these two commits for addition into a future GreaseMonkey release.

Thanks!

John

### DIFF
--- a/components/greasemonkey.js
+++ b/components/greasemonkey.js
@@ -108,7 +108,8 @@ function createSandbox(
     aScript, aContentWin, aChromeWin, aFirebugConsole, aUrl
 ) {
   var sandbox = null;
-  if (GM_util.inArray(aScript.grants, 'none') || aScript.nosandbox) {
+  if (GM_util.inArray(aScript.grants, 'none')
+      || (aScript.nosandbox && aScript.userNosandbox)) {
     // If there is an explicit none grant or nosandbox is set,
     // use a plain unwrapped sandbox with no other content.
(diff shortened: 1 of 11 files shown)


komichi:
Given that the security concerns you raised have been addressed completely, it is difficult to understand why you are closing this without so much as a comment?

arantius:
I'm not comfortable bringing GM_xhr any closer to content than it already is.

PR #1861, opened 2014-01-21
Updated AddonManager userscripts to match other addon-types

jerone:
It always bothered me that the add-on manager for userscripts is a little bit different then the other add-on types. 

So today I updated the:
- _scripts listing_ - Added the following:
  - **More** - Added the `More` link, linking to the scripts detail page.
- _details page_ - Added the following meta:
  - **Creator** - Shows the author when `// @author` is defined in the userscript. When homepage (see below) is defined, creator will be linkable.
  - **Homepage** - Adds link to scripts homepage when `// @homepageURL` is defined in the userscript.
- _context-menu_ - Right-click on an userscript. Added menuitems for:
  - **Show More Information** - Opens the details page. This item is default behavior, so it's bold.
  - **About** - Opens the about dialog.
  - **Execute sooner & later** - Disable second & second to last execution order menuitem. Fixes greasemonkey#1220

No localization updates, as all are from Fx own localization.

Maybe TODO's:
- Add more metadata's that have same attributes in Fx (like `// @supportURL` && `// @contributionURL` and more).
- Add list of all remaining metadata's (e.g. `// @grant`).
- ~~Minimize menuitems `Execute (first|sooner|later|last)` (maybe a sub-menu) -> see comment https://github.com/greasemonkey/greasemonkey/pull/1861#issuecomment-33366513 & https://github.com/greasemonkey/greasemonkey/pull/1861#issuecomment-33430303~~


Ede123:
Hmm, why is this closed? It was never merged, right?

Actually I think this would be a really helpful pull request!

Issue #1914, opened 2014-04-30
"View User Script Source" - Why directly below "Open New Tab"?


hurda:
When right-clicking on a link to a .user.js-file, GM adds an entry to the context-menu to view they script's source. and a menu-seperator.

https://github.com/greasemonkey/greasemonkey/blob/f6c81a7ab013b3a46df5ec926fb699c6e4982e34/content/browser.xul#L30

But when a user is hiding some of the context-menu-entries with a userstyle, there can be two separators right next to each other.

![gm](https://cloud.githubusercontent.com/assets/3539609/2843590/b3f28458-d080-11e3-98fb-71127ce035eb.png)

Wouldn't it more reasonable to place the "View User Script Source" below all "Open Link in ..."-entries? This would render the additional separator unnecessary.

The IDs for them can be found here: http://mxr.mozilla.org/mozilla-central/source/browser/base/content/browser-context.inc?rev=120a6f93c1d4#19

PR #1979, opened 2014-07-30
"View User Script Source" - Why directly below "Open New Tab"?

janekptacijarabaci:
Ad #1914
The suggestion (for example).


janekptacijarabaci:
In the 4.0 version will not make sense.

Issue #1999, opened 2014-08-23
Create New Script with Unicode encoding


TimidScript:
Currently when you create a script it creates it with the default system encoding. If system supports it, shouldn't it create it in unicode.

arantius:
Ok here's my standard character encoding test, for this issue.

1. Search wikipedia for Toyota.
2. Copy the name in Japanese (トヨタ自動車株式会社).
3. Click "new user script".
4. Paste that into the script name field.  (With a few leading and trailing captital-N, making later steps easier.)
5. Close the editor that launches without making any changes/saving.

The file that ends up on my disk:

```
$ grep '@name ' NNN_トヨタ自動車株式会 社_NNN.user.js | hexdump -C
00000000  2f 2f 20 40 6e 61 6d 65  20 20 20 20 20 20 20 20  |// @name        |
00000010  4e 4e 4e 20 e3 83 88 e3  83 a8 e3 82 bf e8 87 aa  |NNN ............|
00000020  e5 8b 95 e8 bb 8a e6 a0  aa e5 bc 8f e4 bc 9a e7  |................|
00000030  a4 be 20 4e 4e 4e 0a                              |.. NNN.|
00000037
```

So the second line starts `4e 4e 4e`, the capital Ns, followed by `20`, the space.  Then: `e3 83 88 e3  83 a8`...

What do those bytes mean?  Well [U+30C8](http://www.fileformat.info/info/unicode/char/30c8/index.htm) is "Katakana letter to" (ト).  In UTF-8 encoding, it's 0xE3 0x83 0x88.  Hey, that's the first three bytes there!  Then we've got [U+30E8](http://www.fileformat.info/info/unicode/char/30e8/index.htm) "Katakana letter yo" (ヨ), which in UTF-8 is ... 0xE3 0x83 0xA8.  Hey, those are the following three bytes!

AFAICT this is working as expected.

PR #2000, opened 2014-08-24
Create New Script with Unicode encoding

janekptacijarabaci:
Ad #1999
The suggestion (for example).

__But do not save the file in Scratchpad editor!__


arantius:
http://stackoverflow.com/a/2223926/91238

"Use of a BOM is neither required nor recommended for UTF-8"

Issue #1945, opened 2014-06-28
HTTP Basic Auth not cached when using GM_xmlhttpRequest


angrier-muffin:
When initially sending the authorization header with GM_xhr, the server properly validates the user/password and returns the protected page.  Subsequent requests, however, do not work and instead require the authorization header be set explicitly for the request to go through.

Is there a reason why the authorization is not cached after the initial request when sent via GM_xhr?

**Non-GM behavior:** First request prompts user/password.  Subsequent requests do not.

**GM behavior:**

*When first request does not contain authorization header and is sent via GM_xhr:*
Firefox's usual user/password appear and subsequent GM_xhr requests do not require authorization header be set explicitly.

*When first request does contain authorization header and is sent via GM_xhr:*
First request goes through fine without Firefox's prompt.  Each subsequent requests requires authorization headers be explicitly set or Firefox will prompt for user/password.

Perhaps I'm not setting the headers correctly or is this intended behavior?  Here's how I'm sending the request:

```
GM_xmlhttpRequest({
	method: 'GET',
	url: 'https://www.website.com/',
	headers: {
		'Authorization':'Basic ' + base64encodedUserAndPassword,
	},
	onload: function(r) {
		// handle response here
	}
});
```

Is there a way to send the initial request via GM's xhr method without having to send the user/password on each subsequent attempt, or do I simply have to send the initial request and prompt to browser's built in handling, and then send my requests like normal after that?

Is the authorization only cached when done through Firefox's built-in dialog?

arantius:
Today, when I run this script:

```javascript
// ==UserScript==
// @name        GM_xhr HTTP Auth Test
// @include     http://localhost/*
// @exclude     http://localhost/auth*
// @version     1
// @grant       GM_xmlhttpRequest
// ==/UserScript==

GM_xmlhttpRequest({
  method: 'GET',
  url: "http://localhost/auth/",
  headers: {
    'Authorization': "Basic dGVzdDp0ZXN0",
  },

  onerror: function() { dump('err!\n'); },
  onload: function(xhr) {
    dump('load 1!\n');
    dump(xhr.responseText.substr(0, 128).replace(/\n/g, ' ')+'\n');

    GM_xmlhttpRequest({
      method: 'GET',
      url: "http://localhost/auth/",
      onerror: function() { dump('err!\n'); },
      onload: function(xhr) {
        dump('load 2!\n');
[...]

PR #2008, opened 2014-08-31
HTTP Basic Auth not cached when using GM_xmlhttpRequest

janekptacijarabaci:
Ad #1945
The suggestion (for example).

The user script:
https://gist.github.com/janekptacijarabaci/495f56ac22bdf25cdc40


janekptacijarabaci:
__Ad worksforme:__

Gist: https://gist.github.com/janekptacijarabaci/495f56ac22bdf25cdc40

GM 3.4.1:
Show login prompt

After applying this fix:
Does not show login prompt

arantius:
See #1945 ;  works for me now, so rejecting the pull request as unnecessary.  (And, wow, complex too.)

PR #2054, opened 2014-11-03
Add missing extractMeta import in newscript.js

Ventero:
Looks like I missed this call to `extractMeta` in 23205fefc4a0cd209eeb2cca8fe5404a78d4fea4. Double-checked now, and this was the only use without import.


arantius:
Sorry I missed this for so long, ended up noticing it separately in 94bdf7b0b094d66c2cd27ef571ec01b57843eb78 .

PR #2057, opened 2014-11-06
GM_log truncates at NULL byte (typeof 'string' only)

janekptacijarabaci:
The suggestion (for example).

The problem - example:
GM_log(GM_listValues());

Browser Console:


arantius:
Already in as 61870329 .

PR #2081, opened 2015-01-29
Update greasemonkey.dtd

ClassicUA:


arantius:
Thanks for this but we manage translations at babelzilla ( http://beta.babelzilla.org/projects/p/greasemonkey/ ).

Issue #2038, opened 2014-10-15
Pale Moon compatibility


Gittyperson:
Greasemonkey up to v.1.15 works perfectly on this popular browser. Is there a possibility we could see the newer versions working with it?

Crazycatz00:
I messed around with getting Greasemonkey 2.3 working in Palemoon 25.1, and I was able to get it to work; but not without disadvantages:
* `createObjectIn`, `cloneInto`, `exportFunction` are not supported. The export code is fine though; it's just exporting `undefined`.
* `Components.utils.getObjectPrincipal` is not supported. This is only used when checking the callbacks for GM_xmlhttpRequest; I only got it to work by checking if the function exists, and if not bypassing the check (which re-introduces #&#8203;1937 for browsers without support. v1.15 doesn't have any check like this though, from what I can see.).

I'm not sure how big of a security problem it would be to selectively bypass the callback check, but if it's not bad I could try to do a fork for this.

DAOWAce:
So it needed more changes than just the GUID?

Well, I hope the scripts I still use don't need anything from GM versions above what works with PM (1.15), as I'm just about to re-transition off my old v28 Firefox install.

I disagree with PM 25's GUID change which broke most addons in the first place (and halted my initial transition over to it all this time), but oh well, development seldom moves backwards and eventually we just all gotta deal with it.  Still worlds better than dealing with Australis..

JamesTheAwesomeDude:
:+1: Thanks for fixing this! I really appreciate Pale Moon support.

koitsu:
An additional :thumbsup: from me as well for Pale Moon support.  Things like the BTTV addon don't work in PM (for different reasons than GM), so being able to get working GM / Userscripts under PM is super helpful.  Thank you so much for your effort on this.

JamesTheAwesomeDude:
Hey, on the Mozilla Add-Ons page, https://addons.mozilla.org/en-us/firefox/addon/greasemonkey/, the Download button looks like it's to download the latest 3.3 version, but it actually downloads version [0.9.22](https://addons.mozilla.org/en-us/firefox/addon/greasemonkey/versions/0.9.22).

[...]

PR #2092, opened 2015-02-25
Pale Moon Support

Crazycatz00:
Add #2038.
I've been personally running this on PM 25.2.1x64 for a couple weeks now without problem, though I haven't tested extensively beyond my normal usage.

### DIFF
--- a/modules/xmlhttprequester.js
+++ b/modules/xmlhttprequester.js
@@ -5,7 +5,10 @@ Components.utils.import("resource://greasemonkey/util.js");
 function GM_xmlhttpRequester(wrappedContentWin, originUrl, sandbox) {
   this.wrappedContentWin = wrappedContentWin;
   this.originUrl = originUrl;
-  this.sandboxPrincipal = Components.utils.getObjectPrincipal(sandbox);
+  // Firefox < 29 does not support getObjectPrincipal in a scriptable context.
+  // Older Greasemonkey didn't use this, so if the browser doesn't support it,
+  // this shouldn't be less secure (for that browser).
+  this.sandboxPrincipal = 'function' == typeof Components.utils.getObjectPrincipal ? Components.utils.getObjectPrincipal(sandbox) : null;
 }
 
 // this function gets called by user scripts in content security scope to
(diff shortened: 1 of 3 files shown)


modules/xmlhttprequester.js
```diff
@@ -169,10 +172,13 @@ function(wrappedContentWin, req, event, details) {
   var eventCallback = details["on" + event];
   if (!eventCallback) return;
 
-  // ... but ensure that the callback came from a script, not content, by
-  // checking that its principal equals that of the sandbox.
-  var callbackPrincipal = Components.utils.getObjectPrincipal(eventCallback);
-  if (!this.sandboxPrincipal.equals(callbackPrincipal)) return;
+  // Firefox < 29 hack; see above.
+  if ('function' == typeof Components.utils.getObjectPrincipal) {
```
arantius: So this just skips implementing a critical security feature?  I'm not super fond of that.

Crazycatz00: Ya, this does skip it, but only if the browser doesn't support it. So for FF >= 29, it should still use the check, but for earlier versions (e.g. Pale Moon, which currently use 1.15 lacking this check) it won't. I see this as a gain, as PM can benefit from the other fixes and changes since, but I can see why you wouldn't want any bypasses for this. I just don't see any other way to check without that function. (I'm not well versed on FF internals though.)

I figured that since it does work, and *shouldn't* affect mainline Firefox, I would at least mention it. In the worst case, I could just maintain a PM-only branch with this disabled until I can get them to implement it, though.

arantius: Good point; if the protection is not already in place due to usage of ancient Greasemonkey version, the impact is quite reduced.

Crazycatz00: Just to update, I have requested for Pale Moon to add support for this function. I don't know any time-frame, but when it's done 93611362c9fe3e10e00e893ca537061bf5033155 should be unneeded.

Skrell: I'm sorry, i'm new to git and would like to use the latest version of greasemonkey with the few fixes you mention here...how can i get an xpi with them?

arantius: I think this is in 3.2beta2

https://addons.mozilla.org/en-US/firefox/addon/greasemonkey/versions/?page=1#version-3.2beta2

Skrell: Why won't mozilla let me download this version?  Or 3.2beta1 for that matter?

Issue #2127, opened 2015-03-25
User Scripts display order not remembered in about:addons


CheriMDixon:
You can sort by Status, Name, or Execution Order, but once you leave about:addons and come back later, your preferred sorting is no longer set. Please remember the user's sort selection.

Thanks! :-)

roger21:
also it would be nice have a drag and drop system to order them (execution order) because the current system (earlier/later) is really tedious when you want to order any new script inside a long list .... well actually i just remembered i just edit the config.xml from now ... so, there's an easy way still

CheriMDixon:
Thanks, janekptacijarabaci!

arantius:
Got fixed by 55820c14de40c8c5bdd575e20f7e628cac2ab278 in 3.4.

PR #2128, opened 2015-03-26
User Scripts display order not remembered in about:addons

janekptacijarabaci:
Ad #2127
The suggestion (for example).


arantius:
Merged by hand.

Issue #2130, opened 2015-03-27
Script install fails if the download url has a query string


gruez:
If the download url for the script has a query string (ie. https://example.com/script.user.js?), greasemonkey will fail to recognize it as a script.

erosman:
No need for new RegExp (makes it easier)
````JavaScript
var gScriptEndingRegexp = /\.user\.js($|\?)/;
````

Since `aOriginURI` is `nsIURI` you can also use
````JavaScript
aOriginURI.path.match(gScriptEndingRegexp)
!aContentURI.path.match(gScriptEndingRegexp)
````

`test` is faster than `match` ie
````JavaScript
gScriptEndingRegexp.test(aOriginURI.path)
!gScriptEndingRegexp.test(aContentURI.path)
````

arantius:
I'm wary of changing anything in this area.  It's common, but not required/exclusively done, that specifically the path name part of the URL is the bit that matches the pattern.  This logic has been in place as is [for ten years](https://github.com/greasemonkey/greasemonkey/blob/6d71e08fad907b6c78ce2106fd4fcdb190bf52d7/src/content/browser.xul#L80) and has worked fine so far.

A known workaround is to append extra unused parameters *at the end* when you know that the link points to a user script.  Conversely if you're (e.g.) a  web based file editor, you can (today) add something to the URL after the filename (containing .user.js) to *suppress* the install dialog.  A feature we should not destroy, either.

erosman:
You can still try my suggestion ie

No need for `new RegExp` (makes it easier, fewer escapes)
````JavaScript
var gScriptEndingRegexp = /\.user\.js$/;
````

Since `aOriginURI` & `aContentURI` are `nsIURI` you can also use
````JavaScript
aOriginURI.path.match(gScriptEndingRegexp)
!aContentURI.path.match(gScriptEndingRegexp)

ulrichb:
Please take this improvement into consideration again. Note that the workaround with the extra parameter doesn't work in my case because I have generated links which I can't influence.

ulrichb:
So... Has this been fixed?

PR #2131, opened 2015-03-27
fix for #2130

gruez:
If the download url for the script has a query string (ie. https://example.com/script.user.js?), greasemonkey will fail to recognize it as a script. This patch fixes it.


jerone:
Then you probably want to add support for fragment identifier by adding the hashtag `#` too...

```diff
-var gScriptEndingRegexp = new RegExp('\\.user\\.js$');
+var gScriptEndingRegexp = new RegExp('\\.user\\.js($|\\?|#)');
```

gruez:
I don't think that's necessary because fragments are nonsensical in a javascript file.

jerone:
I guess you're right; a server can't use it, and a browser can't cache it.

Issue #2132, opened 2015-03-31
3.1: GM_info.version is "unknown"


tiansh:
In Greasemonkey 3.1, GM_info.version returns "unknown".

It seems labeled as TODO in greasemonkey/modules/ipcscript.js:61 ; Hope it will be fixed.

Thanks.

arantius:
Whoops!  Good catch.

PR #2133, opened 2015-03-31
GM_info.version is "unknown", fix typo (description), add (noframes)

janekptacijarabaci:
Ad #2132
The suggestion (for example).

### DIFF
--- a/components/greasemonkey.js
+++ b/components/greasemonkey.js
@@ -27,6 +27,13 @@ var gTmpDir = Components.classes["@mozilla.org/file/directory_service;1"]
     .getService(Components.interfaces.nsIProperties)
     .get("TmpD", Components.interfaces.nsIFile);
 
+var GM_GUID = "{e4a8a97b-f2ed-450b-b12d-ee082ba24781}";
+var gGreasemonkeyVersion = 'unknown';
+Cu.import("resource://gre/modules/AddonManager.jsm");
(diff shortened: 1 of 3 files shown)


components/greasemonkey.js
```diff
@@ -27,6 +27,13 @@ var gTmpDir = Components.classes["@mozilla.org/file/directory_service;1"]
     .getService(Components.interfaces.nsIProperties)
     .get("TmpD", Components.interfaces.nsIFile);
 
+var GM_GUID = "{e4a8a97b-f2ed-450b-b12d-ee082ba24781}";
+var gGreasemonkeyVersion = 'unknown';
+Cu.import("resource://gre/modules/AddonManager.jsm");
+AddonManager.getAddonByID(GM_GUID, function(addon) {
```
jerone: [`AddonManager.getAddonByID` is asynchronous](https://developer.mozilla.org/en-US/Add-ons/Add-on_Manager/AddonManager#getAddonByID%28%29) and mite not be called when `gGreasemonkeyVersion` is being used...

janekptacijarabaci: Unfortunately: https://bugzilla.mozilla.org/show_bug.cgi?id=606876
Any idea how to solve it better?

arantius: This (initialize to unknown, then load async.) is always how it worked.

PR #2137, opened 2015-04-05
Remove "installPolicy shouldProcess()" debug line

foutrelis:
I've noticed this debug message gets printed a lot when viewing YouTube videos or possibly visiting other sites with Flash objects.


arantius:
This got fixed somewhere along the way already.

PR #2157, opened 2015-05-02
Suggestion: GM_xhr - add support for loadstart and loadend events (+ timeout in the upload)

janekptacijarabaci:
See:
http://www.w3.org/TR/XMLHttpRequest/#xmlhttprequest
https://hg.mozilla.org/mozilla-central/file/1e8d30cb367e/dom/base/nsXMLHttpRequest.cpp#l92


janekptacijarabaci:
The better pull request: #2161

Issue #2043, opened 2014-10-27
Feature: Execute scripts on demand


AnasHSulaiman:
The most notable use case is to run bookmarklets like [CSS Layout Debugger](https://gist.github.com/addyosmani/fd3999ea7fce242756b1), [CriticalCSS](https://gist.github.com/PaulKinlan/6284142), SaveFrom.net, and many more others.

I imagine the ideal UX for this is having a Sublime-Text-Command-Palette-like interface: An autocomplete input field that uses fuzzy search to filter installed scripts, and runs the first one in the list by simply hitting enter. And this field can be brought up using a shortcut (like ctrl+shift+p).

It would be event better to specify manual scripts using something like `@run-at user-demand`. This is to avoid running some scripts twice.

ahspw:
I have implemented the UI described above with minimum code changes.

Basically, I've added a floating panel that can be activated using CTRL+SHIFT+SPACE. Inside that panel there is a textbox, where can filter for all registered menu commands on the current window, and execute the selected one by hitting enter.

Now any bookmarklet can be simply converted into a user script that registers its function as a menu command, that can be productively executed on demand.

I will submit a pull request in the near future, once I'm done testing.

PR #2158, opened 2015-05-02
fix issue #2043: Feature: Execute scripts on demand

ahspw:
This adds a floating panel that displays all available menu commands for the current page.
The floating panel can be activated by pressing CTRL+SHIFT+SPACE.

User can filter available menu commands by typing a few characters into the text field, and any menu command whose name contains those characters will be displayed.
The first menu command from the list will be selected and ready to execute upon hitting Enter.
User can navigate the list up and down using arrows while still have the text box focused for further filtering.

The following screenshot shows how the floating panel looks like. Also, it demonstrates the ability to execute Pinboard bookmarklet -which Firefox no longer allows it to run due to a security issue- after turning it into a menu command script.
screenshot

### DIFF
--- a/content/browser.js
+++ b/content/browser.js
@@ -53,6 +53,96 @@ GM_BrowserUI.chromeLoad = function(e) {
 
   // Make sure this is imported at least once, so its internal timer starts.
   Components.utils.import('resource://greasemonkey/stats.js');
+
+  // MenuCommands Panel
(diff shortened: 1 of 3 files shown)


arantius:
I don't like this feature.  Scripts can already register event listeners or menu commands, and most scripts will be incompatible.

Issue #2096, opened 2015-03-03
Support magic TLDs in the @match metadata block


DerpMcDerp:
Extend the `@match` metadata block so that `@match http://*.foo.*/whatever` and/or `@match http://*.foo.tld/whatever` work.

arantius:
This would introduce incompatibilities with other user script engines that support `@match`, and negate some of the security benefits.

DerpMcDerp:
If you're worried about security, you could also support an explicit whitelist of the tlds (e.g. `@match http://*.foo.(net|com|co.uk)`) and have greasemonkey filter the given tlds against the magic tld list.

PR #2169, opened 2015-05-12
Support magic TLDs in the @match metadata block

janekptacijarabaci:
Ad #2096
The suggestion (for example).


arantius:
Sorry if I wasn't clear in the original feature request; I don't like this feature and I don't plan on adding it.  Unless maybe there's *overwhelming* user support?

janekptacijarabaci:
OK. I'll keep it as branch :-)

Issue #2172, opened 2015-05-16
GM_getValue: the returned objects are privileged JS objects


janekptacijarabaci:
See
https://github.com/greasemonkey/greasemonkey/issues/2004#issuecomment-69473107

```
// ==UserScript==
// @name        Greasemonkey - issue - 2172
// @include     https://github.com/*
// @grant       GM_setValue
// @grant       GM_getValue
// ==/UserScript==

var o = {
  n: null,
  b: true,
  i: 100,
  s: "text",
  a: [1, 2],
  o: {k: 1},
  // d: document.body,
}

// <unavailable> - https://bugzilla.mozilla.org/show_bug.cgi?id=1136995

// console.log(o, o.n, o.b, o.i, o.s, o.a, o.o, Object.keys(o));
console.log(JSON.stringify(o), o.n, o.b, o.i, o.s, JSON.stringify(o.a),
            JSON.stringify(o.o), JSON.stringify(Object.keys(o)));

GM_setValue("foo", o);

o = GM_getValue("foo");
// console.log(o, o.n, o.b, o.i, o.s, o.a, o.o, Object.keys(o));
console.log(JSON.stringify(o), o.n, o.b, o.i, o.s, JSON.stringify(o.a),
            JSON.stringify(o.o), JSON.stringify(Object.keys(o)));

GM_setValue("foo", "text");

var r = GM_getValue("foo");
console.log(r);

// WITH e10s ON:

// {"n":null,"b":true,"i":100,"s":"text","a":[1,2],"o":{"k":1}}
// null true 100 text [1,2] {"k":1}
// ["n","b","i","s","a","o"]

// Security wrapper denied access to property k on privileged Javascript object.
// Support for exposing privileged objects to untrusted content via __exposedProps__
// is being gradually removed - use WebIDL bindings or Components.utils.cloneInto
// instead. Note that only the first denied property access from a given global object
// will be reported.

// {}
// undefined undefined undefined undefined undefined undefined
// []

// text

// WITH e10s OFF:

// {"n":null,"b":true,"i":100,"s":"text","a":[1,2],"o":{"k":1}}
// null true 100 text [1,2] {"k":1}
// ["n","b","i","s","a","o"]

// Security wrapper denied access to property k on privileged Javascript object.
// Support for exposing privileged objects to untrusted content via __exposedProps__
// is being gradually removed - use WebIDL bindings or Components.utils.cloneInto
// instead. Note that only the first denied property access from a given global object
[...]

PR #2173, opened 2015-05-17
GM_getValue: the returned objects are privileged JS objects

janekptacijarabaci:
Ad #2172
The suggestion (for example).

Gist: https://gist.github.com/janekptacijarabaci/97e7635cb7f496c005fa

### DIFF
--- a/modules/storageFront.js
+++ b/modules/storageFront.js
@@ -14,9 +14,10 @@ var EXPORTED_SYMBOLS = ['GM_ScriptStorageFront'];
 
 // \\ // \\ // \\ // \\ // \\ // \\ // \\ // \\ // \\ // \\ // \\ // \\ // \\ //
 
-function GM_ScriptStorageFront(aScript, aMessageManager) {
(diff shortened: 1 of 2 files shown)


modules/storageFront.js
```diff
@@ -64,7 +65,11 @@ GM_ScriptStorageFront.prototype.getValue = function(name, defVal) {
   if (value === undefined || value === null) return defVal;
 
   try {
-    return JSON.parse(value);
+    value = JSON.parse(value);
+    // JSON (JavaScript Object Notation) is not designed for serializing
+    // DOM Nodes. But maybe in the future...
```
arantius: What does this comment mean?  When would we ever want a DOM node in context of a storage call?

janekptacijarabaci:
> When would we ever want a DOM node in context of a storage call?

I don't know :-) But I guess it could come in handy. Like many other things...
But feel free to delete it (+ "wrapReflectors: true").

Issue #1933, opened 2014-06-06
CustomEvent constructor throws NS_ERROR_FAILURE in beta Greasemonkey releases (1.16+)


ccd0:
It seems to happen if the script has been granted privileges and the `detail` is set to an object.

For example

```
// ==UserScript==
// @name        testing
// @namespace   testing
// @version     1
// @match       *://example.com/*
// @grant       GM_getValue
// ==/UserScript==

try {
    var x = new CustomEvent("test", {detail: {foo: "bar"}});
} catch (err) {
    console.log(err);
}
```

throws an exception:

```
Exception { message: "", result: 2147500037, name: "NS_ERROR_FAILURE", filename: "", lineNumber: 0, columnNumber: 0, inner: null, data: null }
```

I tested in
Iceweasel 30.0, Firefox 32.0a1
Greasemonkey 1.16beta2, 1.16beta6, 2.0beta1

It does not happen in Greasemonkey 1.15.

Issue #2182, opened 2015-05-25
Pale Moon + GM 3.2beta3: GM_setValue not working


ccd0:
So now that #2092 is merged, I tried testing 3.2beta3 with Pale Moon 25.4.1 (x86 Linux build).  This script

```
// ==UserScript==
// @name        GM_setValue/GM_getValue test
// @namespace   test
// @description Test GM_setValue and GM_getValue.
// @include     http://example.com/
// @version     1
// @grant       GM_setValue
// @grant       GM_getValue
// ==/UserScript==

GM_setValue('foo', 'bar');
alert([typeof GM_getValue('foo'), JSON.stringify(GM_getValue('foo'))]);
```

alerts `object,null` in Pale Moon whereas I get the expected `string,"bar"` in Firefox.

Also alerting @Crazycatz00.  Is anyone volunteering to look into Pale Moon issues, and if so, where should they go?

PR #2183, opened 2015-05-25
Pale Moon + GM 3.2beta3: GM_set/getValue, GM_xmlhttpRequest not working

janekptacijarabaci:
Ad #2182
The suggestion (for example).

PaleMoon: GM_set/getValue - added support for objects

### DIFF
--- a/modules/storageFront.js
+++ b/modules/storageFront.js
@@ -77,7 +77,7 @@ GM_ScriptStorageFront.prototype.getValue = function(name, defVal) {
           value[exposedProps] = {};
         }
         for (prop in value) {
-          if (hasOwnProperty.call(value, prop)) {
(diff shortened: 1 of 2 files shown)


ccd0:
Can confirm this fixes `GM_setValue` and `GM_getValue` in Pale Moon on my end.

By the way, any idea how much work it would take to fix #1933 in Pale Moon?  In Firefox the issue is solved with `cloneInto` but Pale Moon apparently doesn't have that.

arantius:
Sorry.  Pale moon is not supported.  It's not compatible (enough) with Firefox.  All these conditional branches create a significant support burden, which I'm not prepared to assume.

janekptacijarabaci:
The partial and temporary solution to the problem:
https://github.com/janekptacijarabaci/greasemonkey/releases/latest

Issue #2185, opened 2015-05-28
Pale Moon + GM 3.2beta3: GM_openInTab not working 


ccd0:
Testing with 3.2beta3 with Pale Moon 25.4.1 (x86 Linux build) again.  Attempting to use GM_openInTab results in no tab being opened and the following error in the browser console:

```
[02:37:51.872] tabBrowser.getTabForBrowser is not a function @ chrome://greasemonkey/content/browser.js:71
```

which links to this line:
https://github.com/greasemonkey/greasemonkey/blob/7d0f000040aba77c017fbca02c5b539a36c79e04/content/browser.js#L71

arantius:
Since #2038 I have learned that Pale Moon is really truly a *fork* of Firefox, from version 25 ([October, 2013](http://en.wikipedia.org/wiki/History_of_Firefox#Version_25)).  Firefox is moving further and further from what it was then.  To keep working in Firefox, we have to change more and more.

I'm getting pretty convinced that attempting to continue supporting Pale Moon is a losing game.

janekptacijarabaci:
Even from version 24...

ThrawnCA:
"I'm getting pretty convinced that attempting to continue supporting Pale Moon is a losing game."

I hope there's been some good progress on https://github.com/greasemonkey/greasemonkey/issues/2275 then...continuing to support Firefox is going to require work on a larger scale than the compatibility tweaks needed for Pale Moon.

PR #2187, opened 2015-05-28
Pale Moon + GM 3.2beta3: GM_openInTab not working

janekptacijarabaci:
Ad #2185
 The suggestion (for example).


arantius:
Sorry.  Pale moon is not supported.  It's not compatible (enough) with Firefox.  All these conditional branches create a significant support burden, which I'm not prepared to assume.

janekptacijarabaci:
The partial and temporary solution to the problem:
https://github.com/janekptacijarabaci/greasemonkey/releases/latest

Issue #2191, opened 2015-05-30
Script preferences window too large


arantius:
See: https://groups.google.com/forum/#!topic/greasemonkey-users/qu1ywWZo8lo

NoneGiven:
Any word on when this will land? I've long had a batch file to hide/unhide the Windows taskbar that I can access in a few keystrokes, and this STILL proves to be a gigantic headache for me whenever I want to do something with my userscripts (especially back and forth between browsers). I can only imagine how frustrating it is for everyone else with screen resolutions like mine.

PR #2194, opened 2015-06-05
Suggestion: The dialog resizing (for Windows OS), the template is not wrapped, style clean up

janekptacijarabaci:
Ad #2191
The suggestion (for example).

### DIFF
--- a/content/bindings.xml
+++ b/content/bindings.xml
@@ -125,8 +125,10 @@
             this, 'id', 'btnEdit');
         this._btnDel = document.getAnonymousElementByAttribute(
             this, 'id', 'btnDel');
-        if (this == document.getElementById('globalExcludes')) {
-          document.getElementById('resizer').style.display = 'block';
+        if (navigator.appVersion.indexOf("Win") != -1) {
+          if (this == document.getElementById("globalExcludes")) {
+            document.getElementById("resizer").style.display = "block";
+          }
         }
       </constructor>
 
--- a/content/bindings.xml
+++ b/content/bindings.xml
@@ -105,7 +105,7 @@
   >
     <content>
       <xul:hbox flex="1">
-        <xul:listbox flex="1" id="listbox" rows="8" />
+        <xul:listbox flex="1" id="listbox" />
         <xul:vbox>
           <xul:button id="btnAdd" label="&button.add;" oncommand="_onAddPage();" />
           <xul:button id="btnEdit" label="&button.edit;" oncommand="_onEditPage();"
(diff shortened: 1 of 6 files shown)


arantius:
I don't understand.  These grippies are not in the kind of place that makes sense, I can't understand what they're each for.

janekptacijarabaci:
Unfortunately, also I do not understand. What are those arrows? Arrows point to the <resizer> element - this allows you to stretch the window (in this case, horizontally).

arantius:
Sorry I still don't understand this.  The window can be resized just like any other, by its corner.  Long text is truncated.  Why this change?

janekptacijarabaci:
> The window can be resized just like any other, by its corner.

Confirmed on:
Ubuntu 15.04

Not confirmed on:
Windows 7/8.1 :-)
(windows are modal windows - the window resizing doesn't work)

And then I suggested: "The template is not wrapped". What is your opinion on this?

arantius:
I don't know what "the template is not wrapped" means.

janekptacijarabaci:
https://github.com/greasemonkey/greasemonkey/pull/2194/files#diff-bb4d6a2d4a58244882529c2436edfe21R53

I'm sorry for the big "commit". I can do better (resizing the window, new script template, fix typo...). It was just the suggestion (for simplicity).

Now:

vs.
The suggestion:

arantius:
Oh of course that's what you mean by "the template", your earlier screenshots were just overly cropped.  Great.

Also, yes.  Windows won't let you resize this window, but there's no reason to want that behavior.

arantius:
49b1cc8 is only two lines and gets quite close to this, what do you think?

As far as I can tell the only thing it doesn't do is allow resizing when clicking the "options" button in the AOM.

janekptacijarabaci:
It's not occurred to me :-)

> As far as I can tell the only thing it doesn't do is allow resizing when clicking the "options" button in the AOM.

That's what this trick does:
https://bugzilla.mozilla.org/show_bug.cgi?id=267950#c9

```
javascript:opener.openDialog("chrome://greasemonkey/content/options.xul","","modal,resizable");window.close();
```

arantius:
Calling this done, filed #2300 to track the bit that isn't.

jerone:
Maybe force not wrapping lines in the "New User Script" dialog "Includes" and "Excludes" settings. Makes the "One per line" comment clearer.

janekptacijarabaci:
https://github.com/greasemonkey/greasemonkey/pull/2194#issuecomment-148831736: +1

PR #2207, opened 2015-06-26
Fix height of script settings window for Full HD and lower screen sizes

ghost:
I was unable to see full window on my HD resolution screen because of big size of lists in cludes tag. So I decided to add fixed max-height for low resolutions and used 25% of SVGA height.


arantius:
Refs #2191

arantius:
Thanks for your submission; I've gone for a more thorough approach ( eccc46c ).

Issue #2225, opened 2015-07-25
GM 3.3: GM_registerMenuCommand - commandFunc (throws the warning in the Browser Console)


janekptacijarabaci:
![warning](https://cloud.githubusercontent.com/assets/2373486/8888285/b35c3824-32a7-11e5-887e-1cddfe69d276.png)

arantius:
The value isn't used, so the right fix is to not pass it at all.  (Construct a new object that has only the plain string values that are actually used.)

PR #2230, opened 2015-08-04
GM 3.3: GM_registerMenuCommand - commandFunc (throws the warning in the Browser Console)

janekptacijarabaci:
Ad #2225

The suggestion (for example).

__Warning: Thus it will not work in Firefox 24- (e.g. PaleMoon)! This works in Firefox 25+__


janekptacijarabaci:
https://github.com/greasemonkey/greasemonkey/issues/2225#issuecomment-127604085

PR #2233, opened 2015-08-09
GM 3.2+: GM_openInTab(url, false) - open the page in the background

janekptacijarabaci:
``` javascript
// browser.tabs.loadInBackground = true
GM_openInTab(url, false); // open the page in the background
```

The suggestion (for example).


arantius:
I don't understand this change.  In what input condition of `prefBg` and `aMessage.data.inBackground` are the results different before/after it?

janekptacijarabaci:
The parameter ("false") should take precedence over "browser.tabs.loadInBackground" ("true")
= the result should be "false" (but: false OR true (bitwise OR assignment operator "|=") = true)

However, it also solves: #2247

Issue #2237, opened 2015-08-15
Pass sameZoneAs to sandbox constructor to make GCs cheaper


the8472:
According to about:memory greasemonkey sandboxes all seem to live within the same zone (in process/remote tab child global, depending on e10s).

As I understand it this has a negative impact on GC performance since allocation by greasemonkey scripts triggers GCs in this fairly large zone instead of GCing in a window-specific zone.

This can be mitigated by using the [sameZoneAs option](https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XPCOM/Language_Bindings/Components.utils.Sandbox#options).

The addon SDK [does this for its sandboxes](https://github.com/mozilla/addon-sdk/blob/8e6cc79651aae915471913e45a2d38a4d651ac6f/lib/sdk/content/sandbox.js#L150), so it's probably best practice.

arantius:
Pull request merged.

PR #2239, opened 2015-08-17
Optimizations for profiles with many tabs

the8472:
Makes the sandboxes more gc-friendly as mentioned in #2237 and reduces the number of observer registrations.

### DIFF
--- a/content/framescript.js
+++ b/content/framescript.js
@@ -1,11 +1,12 @@
 // The frame script for Electrolysis (e10s) compatible injection.
 //   See: https://developer.mozilla.org/en-US/Firefox/Multiprocess_Firefox
 
+'use strict';
+
 var Cc = Components.classes;
 var Ci = Components.interfaces;
 var Cu = Components.utils;
 
-Cu.import('resource://gre/modules/Services.jsm');
 Cu.import('resource://gre/modules/XPCOMUtils.jsm');
 
 Cu.import('chrome://greasemonkey-modules/content/GM_setClipboard.js');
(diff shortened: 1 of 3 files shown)


arantius:
This pull request appears to do several distinct things, and clash with existing project style.

It would probably be a good idea to split it into two parts to discuss separately.  The one is an easy merge, the other is not.

arantius:
Also FYI I'll be resistant to merging anything that uses language features that [JSDT](http://www.eclipse.org/webtools/jsdt/) fails to understand, as I tend to rely on functionality it provides.  When it can't parse the file, all its features break.  This basically means ES5 and above should be avoided, like the function syntax you've added.

Plus, switching an existing file to strict mode should change independently of anything else.  And I'm not sure how I feel about it yet; if we use it, we should probably be consistent about using it (i.e. everywhere).

the8472:
Ok, I'll split it up and change the syntax.

I'll leave the 'use strict' to you. I think it's just good style since it catches scoping errors more quickly.

arantius:
Thanks for your contribution by the way, as well as working with me to improve them.

janekptacijarabaci:
> I'll leave the 'use strict' to you. I think it's just good style since it catches scoping errors more quickly.

e.g. after installing a script:
js-strict_error

the8472:
btw, have you had a look at [Tern](https://github.com/angelozerr/tern.java/wiki/Tern-&-ECMAScript-support)? it seems to have es6 support

PR #2241, opened 2015-08-17
Use a single shared observer

the8472:


the8472:
is strict mode fine for a completely new file?

modules/documentObserver.js
```diff
@@ -0,0 +1,41 @@
...
+var Cu = Components.utils;
+
+Cu.import('resource://gre/modules/Services.jsm');
+Cu.import('chrome://greasemonkey-modules/content/util.js');
+
+
+var callbacks = new WeakMap()
+
+function OnNewDocument(topWindow, callback) {
+	callbacks.set(topWindow, callback)
```
arantius: This is another memory leak of every window ever created being stored permanently?

arantius:
I don't understand when/how this helps anything?

There's already exactly one frame script per content process right?  And there would be one module imported per child process?

modules/documentObserver.js
the8472: It uses a weakmap, it doesn't hold strongly onto the keys (the windows in this case). So when the window goes away it can collect the callback and then it can collect the framescript in the callback's scope.

the8472:
JSM modules (via `Cu.import`) exist once per process. Frame scripts exist once for every tab.

So this reduces observer registrations from once per tab to once per process.

arantius:
Whoa, tested and confirmed.  But I swear that's not how it used to work.  Or I'm going crazy.

But are adding observers really that expensive that adding this extra layer of indirection is worth it?

the8472:
It's multiplicative. The observer gets fired for frames too. And with a registration from each tab it's iframes-in-current-tab * tabs. It adds a few milliseconds.

Sending the sync message to the parent each time is worse, but one step at a time.

PR #2242, opened 2015-08-17
Disable e10s shims

the8472:
See https://developer.mozilla.org/en-US/Add-ons/Install_Manifests#multiprocessCompatible

It shouldn't affect users on the FF release branch and uncover any remaining compatibility issues for e10s testers.

It worksforme(tm), but that of course is no guarantee.

Is there any better way to test than just visiting pages with a script installed?


arantius:
Unfortunately we have no automated tests.  I have a script ( https://gist.github.com/arantius/1157543 ) which exercises much of the API surface, but must be manually inspected.  So many of our issues are actually integration problems that unit tests wouldn't truly help much anyway.

Cherry picked and tested ok.

Issue #2246, opened 2015-08-19
Fix breakage introduced by disabling e10s shims


arantius:
Re: 4ae802bb1246474d38b0feeb22fbc5c308a37f0d

FYI @the8472 this does at least break menu commands, when opening the monkey menu errors are logged:

    TypeError: contentWindow is null browser.js:270:17
    TypeError: el is null emptyEl.js:4:10

First when opening, second when closing.  Maybe that's all that's broken?

the8472:
> Maybe that's all that's broken?
Who knows. :P

I'll help with fixing fallout, but first I have to get my other PRs in order if that's ok.

arantius:
Absolutely.  Like I said, just FYI.  This issue is just so that it gets handled before 3.4 is final, even if we just revert for now.

Thanks for all your effort!

PR #2248, opened 2015-08-20
fixes greasemonkey/greasemonkey#2246

the8472:
build the dropdown menu asynchronously after requesting the frame URLs
through the message manager.


arantius:
I have one idea I want to investigate (which isn't very promising, but ...).  I'd expect to merge this, unless that idea pans out.  Popping into the 3.4 milestone for tracking, for now.

Issue #2202, opened 2015-06-20
View source not working in Firefox beta 39.0b7


acrobat1:
In  Firefox beta 39.0b7 the view source context (right-click) option doesn't work, it now prompts directly for script installation instead of opening the code in a tab: for instance, on this page https://greasyfork.org/en/scripts/3840-translate-on-google right-click the **Install this script button**, then click **View User Script Source**.
The gm-temp folder is created containing the actual script, but it doesn't get displayed in the browser even when  dragging the script it in a tab.

janekptacijarabaci:
__Confirmed on:__
Firefox 39.0 beta 7
Firefox Aurora 40.0a2 (2015-06-19) [e10s on/off]
Firefox Nightly 41.0a1 (2015-06-20) [e10s on/off]

__Not confirmed on:__
Firefox 38.0.5

And apparently it's the similar bug as https://github.com/greasemonkey/greasemonkey/issues/2178
![1](https://cloud.githubusercontent.com/assets/2373486/8268115/e25d8b22-1778-11e5-8d07-b9b799ec1867.png)

yfdyh000:
Hi @arantius, the seems is not works for the "View User Script Source" menu.

arantius:
Does 3.4beta2 fix this?

https://addons.mozilla.org/firefox/downloads/file/341782/greasemonkey-3.4beta2-fx.xpi?src=devhub

yfdyh000:
Looks works fine for Fx40~43, but the seems breaking for 35.0.1 (our min are 35.0).

arantius:
Launched Firefox 35.0.1, view source works from both context menu and install dialog.  Breaking how?

PR #2252, opened 2015-08-22
[e10s]: newChannelFromURI - it sometimes throws the error

janekptacijarabaci:
Ad #2202
The suggestion (for example).
__It is necessary to test all options (e.g.: correct permissions / the correct principal?)!__

error

nsIOService::NewChannelFromURI is deprecated:
https://hg.mozilla.org/mozilla-central/file/66a72727e1b3/netwerk/base/nsIOService.cpp#l603

Unfortunately, there are no references to the "NewChannelFromURI2" APIs in the documentation...

The short documentation:
https://hg.mozilla.org/mozilla-central/file/66a72727e1b3/netwerk/base/nsIIOService.idl#l68

### DIFF
--- a/modules/remoteScript.js
+++ b/modules/remoteScript.js
@@ -10,6 +10,7 @@ Components.utils.import('chrome://greasemonkey-modules/content/scriptIcon.js');
 Components.utils.import('chrome://greasemonkey-modules/content/util.js');
 
 Components.utils.import("resource://gre/modules/XPCOMUtils.jsm");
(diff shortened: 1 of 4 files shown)


the8472:
have you considered using privileged XHRs instead of constructing channels manually? That might be more forward-portable.

Load flags on the channel can be set like this: `xhr.channel.loadFlags |= Ci.nsIRequest.LOAD_ANONYMOUS | Ci.nsIRequest.LOAD_BYPASS_CACHE`

Issue #2103, opened 2015-03-11
GM_xhr error in multi-process Firefox when referrer is set


kuehlschrank:
Mozilla/5.0 (Windows NT 6.1; Win64; x64; rv:39.0) Gecko/20100101 Firefox/39.0
Greasemonkey 3.0beta1

```javascript
// ==UserScript==
// @name        Test
// @include     *
// @grant       GM_xmlhttpRequest
// ==/UserScript==

	GM_xmlhttpRequest({
		method:  'GET',
		url:     'http://example.com/',
		headers: {'Referer': 'http://example.org/'},
		onload:  function(res) {
			console.log(res.status);
		}
	});
```

Error in console:
Component returned failure code: 0x80004001 (NS_ERROR_NOT_IMPLEMENTED) [nsIObserverService.addObserver]1 Test.user.js:159:0

arantius:
https://bugzilla.mozilla.org/show_bug.cgi?id=1143006

arantius:
I'm gonna throw a try/catch to ignore this in for 3.0 and hope for a real fix in 3.1.

arantius:
https://bugzilla.mozilla.org/show_bug.cgi?id=1108827 ?

arantius:
Work around so that at least the script doesn't completely fail is still in place.  According to https://bugzilla.mozilla.org/show_bug.cgi?id=1143006#c8 this needs to be yet another big rewrite to work in e10s because only the child process knows the value to rewrite, but only the parent process can access this api to write it.  Moving to 3.2.

arantius:
If I run this script
```javascript
// ==UserScript==
// @name        GM_xhr with referer
// @include     *
// @grant       GM_xmlhttpRequest
// ==/UserScript==

GM_xmlhttpRequest({
  method:  'GET',
  url:     'http://localhost/h.php',
  headers: {'Referer': 'http://example.org/'},
  onload:  function(res) {
    console.log(res.responseText);
  }
});
```
Where said php is
```php
<?var_dump($_SERVER['HTTP_REFERER']);?>
```
I get `string(19) "http://example.org/"` in the console.  This is in nightly 42 with E10S on.  Something changed upstream, I guess?

PR #2255, opened 2015-08-23
Support GM_xmlhttpRequest referer override in e10s

the8472:
fixes greasemonkey/greasemonkey#2103

### DIFF
--- a/components/greasemonkey.js
+++ b/components/greasemonkey.js
@@ -16,6 +16,7 @@ Cu.import("chrome://greasemonkey-modules/content/sync.js");
 Cu.import("chrome://greasemonkey-modules/content/util.js");
 Cu.import("resource://gre/modules/Services.jsm");
 Cu.import("resource://gre/modules/XPCOMUtils.jsm");
+Cu.import("chrome://greasemonkey-modules/content/refererSetter.js", {});
 
 
 var gStartupHasRun = false;
--- a/modules/refererSetter.js
+++ b/modules/refererSetter.js
@@ -1,0 +1,28 @@
+'use strict';
+
+const EXPORTED_SYMBOLS = [];
+
+Components.utils.import("resource://gre/modules/Services.jsm");
+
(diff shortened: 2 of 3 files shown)


arantius:
Is this necessary?  As I [mentioned](https://github.com/greasemonkey/greasemonkey/issues/2103#issuecomment-126738239) this works for me with no change.  (Nightly, e10s on.  Also stable, e10s not even an option yet.)

Actually, if I completely remove the non-working-in-child-processes observer stuff it continues to work.  So something has changed upstream I guess?

the8472:
Hrrm, last time I checked it by inspecting the headers in the browser toolbox and it didn't show up in without the fix.

I'll retest again.

the8472:
Actually, have you tested with master or with 3.5beta1, because the async messages things also contains this fix since I needed the request observer in the parent process anyway.

See 0bfa0ef82d276cf63df41585da0ff0142e04c0bc

arantius:
Test procedure from [#2103](https://github.com/greasemonkey/greasemonkey/issues/2103#issuecomment-126738239).

Stable profile, FF41, GM3.5beta2 as built from 666ddc49318ea077277d93382712a49090d424ae .  Works.
Dev profile, FF44 e10s on, git working copy at 666ddc49318ea077277d93382712a49090d424ae .  Works.
Dev profile, FF44 e10s off, git working copy at 666ddc49318ea077277d93382712a49090d424ae .  Works.

In all cases I get the referer specified by the script, which the server is reading from request headers and echoing, logged.  If I remove the script header specification, I get NULL logged instead.

When I merged the broadcast branch, I specifically removed all of the observer/referer setting stuff.  Note that branch [has no modules/refererSetter.js](https://github.com/arantius/greasemonkey/tree/broadcast-script-config/modules) and [xmlhttprequester.js](https://github.com/arantius/greasemonkey/blob/broadcast-script-config/modules/xmlhttprequester.js) has no special referer handling.

PR #2258, opened 2015-08-24
View source - fix for Firefox 38.0.5-

janekptacijarabaci:
Ad https://github.com/greasemonkey/greasemonkey/issues/2202#issuecomment-134398265
The suggestion (for example).

### DIFF
--- a/components/greasemonkey.js
+++ b/components/greasemonkey.js
@@ -74,7 +74,8 @@ function startup(aService) {
   parentMessageManager.addMessageListener(
       'greasemonkey:scripts-for-uuid',
       aService.getScriptsForUuid.bind(aService));
-  parentMessageManager.addMessageListener(
+  var mm = Services.ppmm ? Services.ppmm : globalMessageManager;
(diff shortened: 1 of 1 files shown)


components/greasemonkey.js
```diff
@@ -74,7 +74,8 @@ function startup(aService) {
   parentMessageManager.addMessageListener(
       'greasemonkey:scripts-for-uuid',
       aService.getScriptsForUuid.bind(aService));
-  parentMessageManager.addMessageListener(
+  var mm = Services.ppmm ? Services.ppmm : globalMessageManager;
```
arantius: Have you tested both versions of this?  It took me quite a while (long enough to be embarrassing) that the "parent process" and "global" message managers cannot be used interchangeably.

janekptacijarabaci: Tested versions: 35.0, 38.0, 39.0, 40.0, 43.0a1 (e10s: on/off).
But tested only cursorily...

PR #2259, opened 2015-08-26
avoid sync messages on document load

the8472:
This removes the sync `greasemonkey:scripts-for-url` message from the framescript.

builds on work from PR #2243 and #2255

### DIFF
--- a/modules/ipcscript.js
+++ b/modules/ipcscript.js
@@ -1,6 +1,5 @@
 var EXPORTED_SYMBOLS = ['IPCScript'];
 
-Components.utils.import("chrome://greasemonkey-modules/content/extractMeta.js");
 Components.utils.import("chrome://greasemonkey-modules/content/util.js");
 
 function IPCScript(aScript, addonVersion) {
@@ -17,15 +16,13 @@ function IPCScript(aScript, addonVersion) {
   this.namespace = aScript.namespace;
   this.noframes = aScript.noframes;
   this.runAt = aScript.runAt;
-  this.textContent = aScript.textContent;
   this.uuid = aScript.uuid;
   this.version = aScript.version;
   this.willUpdate = aScript.isRemoteUpdateAllowed();
 
   this.requires = aScript.requires.map(function(req) {
     return {
-      'fileURL': req.fileURL,
-      'textContent': req.textContent
+      'fileURL': req.fileURL
     };
   });
 
@@ -33,20 +30,10 @@ function IPCScript(aScript, addonVersion) {
     return {
       'name': res.name,
       'mimetype': res.mimetype,
-      'textContent': res.textContent,
       'url': GM_util.getUriFromFile(res.file).spec
     };
   });
 };
-
-IPCScript.prototype.__defineGetter__('metaStr',
-function IPCScript_getMetaStr() {
-  if (!this._metaStr) {
-    this._metaStr = extractMeta(this.textContent);
-  }
-
-  return this._metaStr;
-});
 
 IPCScript.prototype.info = function() {
   var resources = {};
@@ -60,8 +47,6 @@ IPCScript.prototype.info = function() {
   return {
     'uuid': this.uuid,
     'version': this.addonVersion,
-    'scriptMetaStr': this.metaStr,
-    'scriptSource': this.textContent,
     'scriptWillUpdate': this.willUpdate,
     'script': {
       'description': this.description,
--- a/modules/ipcscript.js
+++ b/modules/ipcscript.js
@@ -1,17 +1,33 @@
 var EXPORTED_SYMBOLS = ['IPCScript'];
 
 Components.utils.import("chrome://greasemonkey-modules/content/util.js");
+Components.utils.import('chrome://greasemonkey-modules/content/abstractScript.js');
+
+
+// IPCScript class
+
 
 function IPCScript(aScript, addonVersion) {
   this.addonVersion = addonVersion;
+  this.enabled = aScript.enabled;
+  this.needsUninstall = aScript.needsUninstall;
+  this.pendingExec = {};
+  this.pendingExec.length = aScript.pendingExec.length || 0
   this.description = aScript.description;
   this.excludes = aScript.excludes;
+  this.userExcludes = aScript.userExcludes;
   this.fileURL = aScript.fileURL;
   this.grants = aScript.grants;
   this.id = aScript.id;
   this.includes = aScript.includes;
+  this.userIncludes = aScript.userIncludes;
   this.localized = aScript.localized;
-  this.matches = aScript.matches.map(function(m) { return m.pattern; });
+  this.matches = aScript.matches.map(function(m) {
+    return m.pattern;
+  });
+  this.userMatches = aScript.userMatches.map(function(m) {
+    return m.pattern;
+  });
   this.name = aScript.name;
   this.namespace = aScript.namespace;
   this.noframes = aScript.noframes;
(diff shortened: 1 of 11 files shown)


janekptacijarabaci:
The warning:

``` javascript
Services.ppmm
```
(__Services.mm__, __Services.cpmm__)

Not compatible for Firefox 38-
https://hg.mozilla.org/mozilla-central/diff/1b3657fee5a8/toolkit/modules/Services.jsm

arantius:
1) Thanks for all your effort @the8472 
2) Please read http://greasemonkey.github.io/style.html and follow it in your pull requests . (I see lots of missing spaces and braces, and the general style is to prefix function arguments' names with an "a".)
3) I'd find it much easier to review/understand pull requests if they were independent; this seems to be a mix of some other ones, plus further changes.

What's the overall approach here?  Effectively the whole config of all scripts and their include/exclude data is stored in every child all the time?  And the fileXhr trick means that we don't have to duplicate the script itself?

the8472:
> The Warning

ok, will fix that

> style

Is there some auto-formatter I could apply? That's more a job for a machine than a human.

> pull requests review

Once you looked at the other two you only need to review f21e40a

> Effectively the whole config of all scripts and their include/exclude data is stored in every child all the time? And the fileXhr trick means that we don't have to duplicate the script itself?

Yes. In each child process, not in each frame.

Additionally, the old code used the sync messages to check for script refreshes, this now piggybacks on http observers which carry similar information to the sync messages anyway. We can extract all the necessary information from requests with `TYPE_DOCUMENT`/`TYPE_SUBDOCUMENT` in their `loadInfo`.

janekptacijarabaci:
The warning:

``` javascript
initialProcessData
```
Not compatible for Firefox 42- (https://bugzilla.mozilla.org/show_bug.cgi?id=1162838 , but: 41.0b6 - in this version does not work).

modules/ipcscript.js
```diff
@@ -1,31 +1,40 @@
...
-Components.utils.import("chrome://greasemonkey-modules/content/extractMeta.js");
 Components.utils.import("chrome://greasemonkey-modules/content/util.js");
+Components.utils.import('chrome://greasemonkey-modules/content/abstractScript.js');
+
+
+// IPCScript class
+
 
 function IPCScript(aScript, addonVersion) {
   this.addonVersion = addonVersion;
```
janekptacijarabaci: It will be undefined...

modules/ipcscript.js
```diff
@@ -1,31 +1,40 @@
...
   this.excludes = aScript.excludes;
+  this.userExcludes = aScript.userExcludes;
   this.fileURL = aScript.fileURL;
   this.grants = aScript.grants;
   this.id = aScript.id;
   this.includes = aScript.includes;
+  this.userIncludes = aScript.userIncludes;
   this.localized = aScript.localized;
   this.matches = aScript.matches.map(function(m) { return m.pattern; });
+  this.userMatches = aScript.matches.map(function(m) { return m.pattern; });
```
janekptacijarabaci: I guess it should be: matches => userMatches

modules/ipcscript.js
```diff
@@ -60,8 +123,6 @@ IPCScript.prototype.info = function() {
   return {
     'uuid': this.uuid,
     'version': this.addonVersion,
```
janekptacijarabaci: I think it will be undefined...

the8472: what is "it"?

modules/ipcscript.js
the8472: yes, thanks

modules/ipcscript.js
janekptacijarabaci: addonVersion => GM_info.version
I think it will be undefined...

the8472: I did not change this code. `this.addonVersion` is [set in the constructor](https://github.com/the8472/greasemonkey/blob/9c9e5af5ceb000f093108f9f4ac6fd1529987eab/modules/ipcscript.js#L11)

modules/sandbox.js
```diff
@@ -121,19 +123,70 @@ function createSandbox(aScript, aContentWin, aUrl, aFrameScope) {
...
+  );
+  
+  // meta depends on content, so we need a lazy one here too
+  Object.defineProperty(waivedInfo,
+    'scriptMetaStr',
+    {get: Cu.exportFunction(
+      function() {
+        var meta = fileCache.get("meta");
+        if(meta === undefined) {
+          meta = extractMeta(this.scriptSource);
```
janekptacijarabaci: It throws the error:

modules/ipcscript.js
janekptacijarabaci: Yes, but:
https://github.com/greasemonkey/greasemonkey/pull/2259/files#diff-e07419236596eaa414a5954c7c90c385R151

However, this does not help:
`return new IPCScript(script, gGreasemonkeyVersion);`

GM_info.version = 'unknown';

the8472: Ok, thanks for digging, i'll have a look at it.

the8472:
@janekptacijarabaci thanks for testing, I think those issues should be fixed now.

janekptacijarabaci:
You're welcome.

However, I see another problem...

After editing the script in Scratchpad and saving - properties of the object (GM_info) is not valid (is not changed):
https://github.com/greasemonkey/greasemonkey/pull/2259/files#diff-9f06c50d889eb78e6776c4b4fc7314cbL66

the8472:
@janekptacijarabaci when I save + refresh the page `GM_info.script` does update accordingly for me

demo: https://i.pantsu.cat/lemhtw.webm

It's pretty much the same old logic, just that it infers a page load from a HTTP observer instead of relying on sync messages from the frame script: https://github.com/the8472/greasemonkey/blob/f21e40a74147e409231e39b08f9387e55cc0cb07/modules/refererSetter.js#L33-L53

janekptacijarabaci:
I'm sorry, but I do not work for me.
Firefox Nightly 43.0a1 (2015-09-14)
Firefox 40.0.3

Steps to reproduce:
1. I trying make a new profile of Firefox (but in the default profile too)
2. Installing Greasemonkey (your version)
3. Installation of this script: https://gist.github.com/janekptacijarabaci/e4a939351e5792d3ba90
4. Open Browser Console
5. Open page github.com in browser - only one page is open (one tab)
5a. In Browser Console: e.g. GM_info.script.version: 1.0
6. Open Scratchpad, the script editing (@version 1.0 => @version 1.1), saving
7. Refresh github.com in browser
7a. In Browser Console: GM_info.script.version: 1.0

janekptacijarabaci:
__when="any": this value does not exist...__

Fix this bug (for example):

* https://github.com/greasemonkey/greasemonkey/pull/2259/files#diff-9be6aefe99ccf1e149776242db5940a6R49

From:
```javascript
.scriptRefresh("any", channel.URI.spec, windowId, browser);
```
To:
```javascript
.scriptRefresh(channel.URI.spec, windowId, browser);
```
* https://github.com/greasemonkey/greasemonkey/pull/2259/files#diff-e07419236596eaa414a5954c7c90c385R193

From:
```javascript
service.prototype.scriptRefresh = function(when, url, windowId, browser) {
```
To:
```javascript
service.prototype.scriptRefresh = function(url, windowId, browser) {
```

* https://github.com/greasemonkey/greasemonkey/pull/2259/files#diff-e07419236596eaa414a5954c7c90c385R199

From:
```javascript
this.config.updateModifiedScripts(when, url, windowId, browser);
```
To:
```javascript
this.config.updateModifiedScripts("document-start", url, windowId, browser);
this.config.updateModifiedScripts("document-end", url, windowId, browser);
```

the8472:
ok, incorporated your changes and rebased on master

arantius:
I've started merging this down.  It's a huge change, I want to give it some time to bake before I commit it to master.

https://github.com/greasemonkey/greasemonkey/compare/master...arantius:broadcast-script-config

janekptacijarabaci:
@arantius:
In case you missed this file "modules/refererSetter.js" (https://github.com/greasemonkey/greasemonkey/pull/2259/files#diff-e07419236596eaa414a5954c7c90c385R104):
https://github.com/greasemonkey/greasemonkey/pull/2259#issuecomment-139967584
https://github.com/greasemonkey/greasemonkey/pull/2259#issuecomment-140209171

arantius:
Point one: no, intentional, merge conflict with c274845d95b6d88d70c8b99f4132cf5e3daa0f1f .

Point two: good catch!

arantius:
See #2289 -- doesn't work with e10s off.  Looking into why now.

arantius:
Also not working in FF 44 with e10s off, so e10s off appears to be the real issue.

the8472:
Odd, I did test with e10s off, maybe I skipped that during the recent rebase.

Anyway, how do you want fixes? New commits on this PR? Rebase? Or against your own branch?

arantius:
Gimme 5, I might be near a "fix".  Something at appears to work better, assuming I understand the operating model.

arantius:
I think 73736edf03220f4dd1fa4224ae65dfaa15c845b4 is the right fix for #2289 and some related issues.

the8472:
Hum, was that some merge conflict? The broadcast shouldn't have been in an `else` block. It doesn't exist in my [branch](https://github.com/the8472/greasemonkey/blob/70fab3eca06006122039403cb190228d2fd872c3/components/greasemonkey.js#L177-L183)

arantius:
Indeed, must have been my fault somehow.  Whoops, [there it is](https://github.com/arantius/greasemonkey/commit/726cbba4ea41e2fea317de1d5b7cf0f4b4809ae3#diff-e07419236596eaa414a5954c7c90c385R167).

the8472:
>In case you missed this file "modules/refererSetter.js" 
> Point one: no, intentional, merge conflict with c274845 .

What about the [script refresh part](https://github.com/the8472/greasemonkey/blob/async-messages2/modules/refererSetter.js#L29-L49)?

arantius:
I didn't know it existed because it was in a file called `refererSetter.js`.  Add it back in a place that makes a bit more sense please?

janekptacijarabaci:
The reminder (a typo):
https://github.com/greasemonkey/greasemonkey/pull/2259#issuecomment-142770673

janekptacijarabaci:
The suggestion:
https://github.com/arantius/greasemonkey/pull/3

arantius:
Merged this into master.

Issue #2151, opened 2015-04-22
Address "unsafe CPOW usage" warnings


arantius:
I'm seeing them now for:

Line 266 and 270 of `browser.js`, in:

    function GM_showPopup(aEvent) {
      function urlsOfAllFrames(contentWindow) {
        var urls = [contentWindow.location.href];
        function collect(contentWindow) {
          urls = urls.concat(urlsOfAllFrames(contentWindow));
        }
        Array.prototype.slice.call(contentWindow.frames).forEach(collect);
        return urls;
      }
        
The `var urls = ...` and the `Array.prototype.slice...` lines.  Line 17, 23, and 28 of `windowId.js`.

Maybe others I haven't noticed yet.

janekptacijarabaci:
Current:
![cpow](https://cloud.githubusercontent.com/assets/2373486/7550145/e58a340a-f656-11e4-96aa-e04a17791dad.png)

janekptacijarabaci:
I've pushed a fix to the testing branch: https://github.com/greasemonkey/greasemonkey/compare/master...janekptacijarabaci:_testBranch_unsafe_CPOW_usage

The suggestion (for example).
__It is necessary to test all options!__

__Warning: Thus it will not work in Firefox 24- (e.g. PaleMoon)! This works in Firefox 25+__

arantius:
Merged by hand.

arantius:
This should be in 3.5beta1; let me know what you think:
https://addons.mozilla.org/firefox/downloads/file/354560/greasemonkey-3.5beta1-fx.xpi?src=devhub

PR #2263, opened 2015-08-31
Address "unsafe CPOW usage" warnings - newscript

janekptacijarabaci:
Ad #2151 (and the commit https://github.com/greasemonkey/greasemonkey/commit/6b05c4803c96ef4f947a229679e2160be13577af)
The suggestion (for example).

Current (nothing else I spotted):

warning


janekptacijarabaci:
See #2282

PR #2270, opened 2015-09-12
GM 3.3beta2+: Fix synchronization of commands between popups

janekptacijarabaci:
The suggestion (for example).

Fix synchronization of commands between popups (and after a page refresh F5 - if the number of commands change: 0, >0).

1st step:

2st step:


janekptacijarabaci:
See https://github.com/greasemonkey/greasemonkey/pull/2277

arantius:
This should be in 3.5beta1; let me know what you think:
https://addons.mozilla.org/firefox/downloads/file/354560/greasemonkey-3.5beta1-fx.xpi?src=devhub

Issue #2280, opened 2015-09-18
(in e10s window) Greasemonkey doesn't intercept some userscript links in GitHub (by presenting the installation dialog), though they end with 'user.js'


darkred:
(using FF Nightly 43.0a1 x64, with GM 3.4.1, in a e10s window)

For example: 
the link that is the `Raw` button in [this page](https://github.com/jerone/UserScripts/blob/63b0f7fd9818c780d05635c9f5acb26afe2d05ef/Github_News_Feed_Filter/Github_News_Feed_Filter.user.js) 

The URL is:
`https://github.com/jerone/UserScripts/raw/63b0f7fd9818c780d05635c9f5acb26afe2d05ef/Github_News_Feed_Filter/Github_News_Feed_Filter.user.js`
_The same link, linkified [here](https://github.com/jerone/UserScripts/raw/63b0f7fd9818c780d05635c9f5acb26afe2d05ef/Github_News_Feed_Filter/Github_News_Feed_Filter.user.js) is intercepted ok._

On the other hand, in a non e10s window (and in FF40 w GM 3.4.1) the link is intercepted ok.

***

For reference, for that same script (in a e10s window), the installation gets intercepted ok 
if you either press the `Install` button in [the script's main page(master branch)](https://github.com/jerone/UserScripts/tree/master/Github_News_Feed_Filter) 
(or if you press the `Github_News_Feed_Filter.user.js` link and then the `Raw` button in that link).

janekptacijarabaci:
See https://github.com/greasemonkey/greasemonkey/blob/3.4.1/modules/installPolicy.js#L82

1)
https://github.com/jerone/UserScripts/blob/master/Github_News_Feed_Filter/Github_News_Feed_Filter.user.js
=>
https://github.com/jerone/UserScripts/raw/master/Github_News_Feed_Filter/Github_News_Feed_Filter.user.js
(https://raw.githubusercontent.com/jerone/UserScripts/master/Github_News_Feed_Filter/Github_News_Feed_Filter.user.js)
__aOriginURI.spec=__
`https://github.com/jerone/UserScripts/blob/master/Github_News_Feed_Filter/Github_News_Feed_Filter.user.js`

2)
https://github.com/jerone/UserScripts/tree/master/Github_News_Feed_Filter
=>
https://github.com/jerone/UserScripts/blob/master/Github_News_Feed_Filter/Github_News_Feed_Filter.user.js
=>
[...]

PR #2283, opened 2015-09-21
[e10s on] The install dialog for *.user.js URLs

janekptacijarabaci:
Ad #2280
The suggestion (for example).

But I don't know if it's the best solution for it...


arantius:
Sorry, could you pull and fix merge conflicts?  I tried to merge, but maybe I got the conflict resolution wrong, the fix didn't seem to work for me.

janekptacijarabaci:
OK. See https://github.com/greasemonkey/greasemonkey/pull/2290

PR #2290, opened 2015-09-24
[e10s on] The install dialog for *.user.js URLs

janekptacijarabaci:
Ad #2280
The suggestion (for example).


arantius:
3ea4de9308b05b542e7c1711dc6473b7997133f9

PR #2307, opened 2015-10-25
Minor fix - remove unused function

janekptacijarabaci:
The suggestion.

Or missing import?
```
Components.utils.import("resource://gre/modules/NetUtil.jsm");
```
If it can be used in the future...


the8472:
yeah, that was some leftover from my refactoring.

looks good to me.

PR #2323, opened 2015-11-02
Update greasemonkey.dtd

Watilin:
A message I came across for the first time afters years of using Greasemonkey.
« Télécharger l'erreur » means “download the error” and I bet this is not what the user wants to do :P


arantius:
Thanks for your contribution but please submit translations via http://beta.babelzilla.org/projects/p/greasemonkey/ (otherwise they'll be overwritten next time we update translations, from there).

Watilin:
Got it!

PR #2333, opened 2015-11-13
in-process cache for frequent calls to GM_getValue()

the8472:
I'm dogfooding this right now, as caching is always tricky. Caching in concurrent environments even more so. But as far as I can tell it seems to work.

Anyway, opinions and additional testers welcome.

---

microbenchmark: https://gist.github.com/the8472/4f5efa0f1f968b738706

measures milliseconds per 10k GM_getValue calls in e10s:

vanilla:

3243.7949999999996
1726.7300000000005
1699.875
1930.4450000000006
2399.8150000000005
1915.4300000000003
1695.7250000000004
2594.84
1842.0200000000004
1682.0249999999978

patched:

299.3200000000006 
327.1349999999993 
326.3900000000003 
325.3650000000007 
331.84500000000025
291.6700000000001 
324.14999999999964
324.1100000000006 
326.25
343.8200000000006

Additionally there might be more indirect performance gains by the parent process not having to service those sync messages by reading from the sqlite DB on the main thread.


the8472:
I've been testing this for a while now with a userscript that runs on multiple tabs and makes regular use of `getValue` / `setValue` and have not observed any issues.

Issue #2306, opened 2015-10-24
Sorting by namespace


dogancelik:
It would be nice if a user could sort their scripts by namespace (this does not change execution order) so they can see the all scripts by same author without scrolling up and down the page.

janekptacijarabaci:
See also https://github.com/greasemonkey/greasemonkey/issues/2145, https://github.com/greasemonkey/greasemonkey/issues/2090

dogancelik:
I want to clarify my issue is about `about:addons` page.

PR #2334, opened 2015-11-14
Sorting by namespace

janekptacijarabaci:
Ad #2306
The suggestion (for example).


arantius:
Namespace is a relatively invisible thing to most users; it's only visible when editing the source of a script.  I'm not sure it's a good idea to make it prominent in the UI, it would probably be confusing.

Issue #2229, opened 2015-07-31
about:blank, the script with alert function - after the restart, the browser hangs


janekptacijarabaci:
Firefox 39.0 - Nightly 42.0a1 (2015-07-31)
GM 3.3

This user script (@run-at __document-end/document-start__):

``` javascript
// ==UserScript==
// @name        Greasemonkey - issue - 2229
// @include     about:blank
// @run-at      document-end
// @grant       none
// ==/UserScript==

alert("hello"); // modal window
```
...after restart of the browser - e.g. after installing a new extension, e.g. using drag and drop (if the page loaded is the "about:blank" and enabled browser history), the browser hangs and has to be killed in the Task Manager (and the script must be manually disabled)!

Refs https://github.com/greasemonkey/greasemonkey/issues/2041#issuecomment-107042627, https://github.com/greasemonkey/greasemonkey/issues/2041#issuecomment-107127991, https://github.com/greasemonkey/greasemonkey/issues/2041#issuecomment-126720515

arantius:
Confirmed, with error:

    JavaScript error: resource://gre/components/nsPrompter.js, line 357: NS_ERROR_NOT_AVAILABLE: Cannot call openModalWindow on a hidden window

Somehow we're catching the (content) window loading while the (chrome) window is still invisible, in the session restore path?

arantius:
I found [where this error is thrown](http://mxr.mozilla.org/mozilla-central/source/toolkit/components/prompts/src/nsPrompter.js#356) but if I try to detect this case (invisible window) the same way, it causes Firefox to crash (!).

arantius:
https://bugzilla.mozilla.org/show_bug.cgi?id=1196270

janekptacijarabaci:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1196270#c3

I've pushed a fix to the testing branch:
https://github.com/janekptacijarabaci/greasemonkey/compare/master...janekptacijarabaci:_testBranch_aboutBlank_browserHangs
...but: isWindowVisible is always true (send __Sync__ Message)

janekptacijarabaci:
RESOLVED WORKSFORME
https://bugzilla.mozilla.org/show_bug.cgi?id=1196270#c5 :-/

janekptacijarabaci:
@arantius 
[...]

PR #2357, opened 2016-01-19
about:blank, the script with alert function - after the restart, the browser hangs

janekptacijarabaci:
Ad #2229 (see [Bug 1196270 - Comment 3](https://bugzilla.mozilla.org/show_bug.cgi?id=1196270#c3))
The suggestion (for example).

### DIFF
--- a/content/browser.js
+++ b/content/browser.js
@@ -20,6 +20,22 @@ GM_BrowserUI.init = function() {
         var href = aMessage.data.href;
         GM_BrowserUI.checkDisabledScriptNavigation(contentType, href);
       });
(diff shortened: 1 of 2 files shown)


the8472:
If this only concerns the edge-case of GM scripts specifically targeting about:blank then sending a sync message for every DOMContentLoaded event is quite some overkill, especially considering that background tabs during session restore also execute frame scripts.

janekptacijarabaci:
They should fix it. But... The best solution is probably nothing :-)

janekptacijarabaci:
See https://github.com/greasemonkey/greasemonkey/issues/2229#issuecomment-241197193

Issue #1895, opened 2014-03-17
No obvious (UI-)indicator for scripts which will not or cannot receive automatic updates


hurda:
Currently, scripts which are either modified or don't have the necessary information to be automatically updated give no visual clue that they won't be considered when scanning for updates.

The only clue is "Find updates" being disabled in each script's context-menu.
For disabled scripts this clue doesn't exist at all, as it isn't available for any of them.

So maybe their background-color can be changed (for example like the design for the Java-plugin in Plugins) or something similar, which would immediately give the information regarding the updating-state of each script.

arantius:
@janekptacijarabaci 's above pull request will be merged for version 3.7.  It's not obvious yet, but at least it's something.

PR #2358, opened 2016-01-20
No obvious (UI-)indicator for scripts which will not or cannot receive automatic updates

janekptacijarabaci:
Ad #1895
The suggestion (for example).


arantius:
That's a very subtle indication, but it's better than nothing.  I'd like to customize the text, but it's set by XBL, which isn't easy to customize.

Merged.

arantius:
This should be fixed in 3.7beta2: https://addons.mozilla.org/firefox/downloads/file/388504/greasemonkey-3.7beta2-fx.xpi?src=devhub

I'd appreciate confirmation either way.

Issue #2371, opened 2016-03-05
Script with @run-at document-start will be executed twice


Infocatcher:
Testcase: https://gist.github.com/Infocatcher/1243f295ca93c94160ec
Install: https://gist.githubusercontent.com/Infocatcher/1243f295ca93c94160ec/raw/document-start.user.js
(and then open about:blank?UserScripts/test - should appear only one `+`)

```js
// ==UserScript==
// @name        about:blank?UserScripts/test
// @namespace   dev/null
// @include     about:blank?UserScripts/test
// @run-at      document-start
// @version     1
// @grant       none
// ==/UserScript==

console.log("test: " + document.readyState);
document.body.appendChild(document.createTextNode("+"));
```

Tested with Greasemonkey 3.7, 3.8beta2 (Firefox 44.0.2).

arantius:
Fix merged.

arantius:
This should be fixed in build 3.8beta3; confirmation would be helpful.

https://addons.mozilla.org/en-US/firefox/addon/greasemonkey/versions/3.8beta3

Infocatcher:
Confirmed, looks fixed for me with GM 3.8beta3 too.
Here I've tested with custom build from master branch: https://github.com/greasemonkey/greasemonkey/pull/2372#issuecomment-211471368
Thanks again. :)

Issue #1820, opened 2013-10-28
Scripts don't always fire on about:blank


nachoCheeseDoritos:
Scripts don't always fire on about:blank.

For instance, if you have a simple script that has the following code (removed unnecessary parts):

// @include about:blank#test
alert('test');

And you go directly to the page by typing the URL in the address bar, it will fire as expected.  If you do any of the following, it will NOT fire:

window.location = 'about:blank#test';
window.open('about:blank#test');
GM_openInTab('about:blank#test');
Have a bookmarked version of it and middle click (opening in a new tab)
--Clicking directly on the bookmark will fire as expected though

In order to get it to fire in these cases, one must manually refresh the page.

PR #2372, opened 2016-03-05
Script with @include about:blank?query (or #fragment) will be executed twice

janekptacijarabaci:
Ad #2371 
The suggestion (for example).

See also https://github.com/greasemonkey/greasemonkey/issues/1820


Infocatcher:
Thanks! Testcase from #2371 now works fine for me.

arantius:
This should be fixed in build 3.8beta3; confirmation would be helpful.

https://addons.mozilla.org/en-US/firefox/addon/greasemonkey/versions/3.8beta3

PR #2391, opened 2016-04-23
no issue: declare var as local

crackevil:


janekptacijarabaci:
It's the same thing here:
https://github.com/greasemonkey/greasemonkey/blob/3.7/modules/sync.js#L283

crackevil:
Some more commits to merge. Anyone can help to review it for me?Thanks

the8472:
`var` in loop headers is not local to the iteration, only `let` is.

http://stackoverflow.com/questions/18465211/javascript-loop-variable-scope

modules/prefmanager.js
```diff
@@ -127,6 +127,7 @@ GM_PrefManager.prototype.remove = function(prefName) {
  * call a function whenever the named preference subtree changes
  */
 GM_PrefManager.prototype.watch = function(prefName, watcher) {
+  if (this.observers[watcher]) return;
```
arantius: It's not unreasonable to write a watcher that is capable of observing several preferences, is it not?

arantius:
I've cherry picked 7239e5506aebedf8848139801e64fe91db4bba1a and 3d3b9debc93a179894586e7ea9da074f17bc914b .  Sorry for the delay.

If there's something left which you'd still like to include, please create one new pull request per logical change and it will be considered.

PR #2417, opened 2016-08-06
If Greasemonkey is disabled, some scripts works (about:blank / about:reader)

janekptacijarabaci:
If Greasemonkey is disabled, some scripts works (if the @include parameter has the value "about:blank" or about:reader...).


janekptacijarabaci:
vs.

Issue #2396, opened 2016-05-05
Greasemonkey API documentation is outdated


CoolCmd:
https://github.com/greasemonkey/greasemonkey/issues/2199
https://wiki.greasespot.net/GM_info

janekptacijarabaci:
I would like to change name...

Also:

https://wiki.greasespot.net/GM_info
vs.
`uuid`, `scriptSource`, `localizedDescription`, `localizedName`, `noframes`, ~~unwrap~~

https://wiki.greasespot.net/GM_openInTab
vs.
https://github.com/greasemonkey/greasemonkey/pull/2257 (synchronize API with Tampermonkey)

https://wiki.greasespot.net/GM_getValue, https://wiki.greasespot.net/GM_setValue
vs.
https://github.com/greasemonkey/greasemonkey/issues/2172 (JSON/objects)

https://wiki.greasespot.net/Metadata_Block
vs.
`author`, `installURL`, `homepageURL`

https://wiki.greasespot.net/GM_xmlhttpRequest
vs.
https://github.com/greasemonkey/greasemonkey/commit/59f07764b8c975cc3cef451743e3f2754fc95f45 (add the `responseXML`)
vs.
https://github.com/greasemonkey/greasemonkey/commit/f6c624d7d4e8b09bd5c30784126fa52a2cffec0d (add the `onloadend`/`onloadstart`)
vs.
https://github.com/greasemonkey/greasemonkey/issues/2330 (add the `mozAnon`/`anonymous`)

...and maybe something else.

arantius:
1) The documentation is on a wiki.  Just register (and perhaps ping me with your username .. fighting the spam war is a pain) and you can fix it yourself.
2) No you can't log in with github.  It's standard mediawiki.  All it needs is a username and password.
3) Thanks for your effort @janekptacijarabaci but deciphering those terse notes is a lot of work, so I have not tried to address any of those.  Maybe some time.

marsjaninzmarsa:
@arantius with pleasure, but registration seems to be broken.

PR #2418, opened 2016-08-06
Tooltip improvements

janekptacijarabaci:
1) Commands: Add localized script name (instead script name)
2) The tooltip (+ list of scripts): Add information about setting `@noscript`

### DIFF
--- a/content/browser.js
+++ b/content/browser.js
@@ -329,13 +329,16 @@ function getScripts() {
     if (script.needsUninstall) return;
     var mi = document.createElement("menuitem");
     mi.setAttribute("label", script.localized.name);
+    if (script.noframes) {
+      mi.setAttribute("tooltiptext", "noframes");
+    }
     mi.script = script;
     mi.setAttribute("type", "checkbox");
     mi.setAttribute("checked", script.enabled.toString());
     if (!noInsert) {
       point.parentNode.insertBefore(mi, point.nextSibling);
     }
-    return mi;
+    return {"menuItem": mi, "noframes": script.noframes};
   }
   getScripts.appendScriptAfter = appendScriptAfter;
 }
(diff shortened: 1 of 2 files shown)


darkred:
> 2) The tooltip (+ list of scripts): Add information about setting `@noscript`

Of course you meant `@noframes` _(there's no `@noscript`)_.

<br/>
Could someone please explain what the values `Active (frames) x/y/z` in the tooltip mean?
The relevant row in browser.js is [485-486](https://github.com/greasemonkey/greasemonkey/blob/master/content/browser.js#L485), i.e.
```
      var activeFrames = runsFramedNoframesDisable + "/"
                         + runsFramedEnable + "/" + runsFramed.length;
```

I'm trying to understand, and 
as I see, the first two (`runsFramedNoframesDisable ` and `runsFramedEnable `) only take the values 0 or 1.
And, if I get this right, all 3 values are for when `// @noframes` is **not** present, and then:
- x (`runsFramedNoframesDisable`) is 1 if there are no nested frames in the document,
- y (`runsFramedEnable `) is 1 if there are nested frames in the document,
- z (`runsFramed.length`) is the count of nested frames

?



----
For reference I quote from [here](https://wiki.greasespot.net/Metadata_Block#.40noframes):
> `// @noframes` metadata imperative:
>
> When present, this imperative restricts the execution of the script. The script will run only in the top-level document, never in nested frames. It takes no arguments, it is either present or not present. This is off (scripts run in frames) by default.

janekptacijarabaci:
> Of course you meant `@noframes` (there's no `@noscript`).

Yes. Thank you :-)

> Active (frames) x/y/z

x - is the count of __active__ scripts in (nested) frames - according to url (without option @noframes)
y - is the count of __active__ scripts in (nested) frames - according to url
z - is the count of scripts in (nested) frames - according to url

e.g.: http://www.w3schools.com/tags/tryit.asp?filename=tryhtml_iframe

It should be in the documentation (in the future)...
https://github.com/greasemonkey/greasemonkey/issues/2396#issuecomment-217111526

darkred:
Thanks a lot for your reply

PR #2420, opened 2016-08-13
WIP: provide APIs to bypass CSP restrictions

the8472:
Not finished yet. The question is, would you accept the following set of new features?

* `@grant baseDomain`. this would use a subsuming principal for the sandbox, e.g. `mail.google.com` -> `google.com`. The principal could then use the same OriginAttributes (i.e. containers, 1st party siloing etc.) but ignore CSP
* `@grant isolateFetch`. add sandbox specific XHR and `fetch` constructors. They would still respect SOP/CORS but when combined with the baseDomain grant they would not be subject to CSP. I.e. they would be less powerful than GM_XHR but more powerful than the page's XHR.
* `@grant GM_blobURI`. This would allow the user to generate privileged URIs which bypass CSP. 

I.e. it could do something like this without running afoul of img-src CSPs:

```javascript
// ==UserScript==
// @name        gm_blob test
// @namespace   test
// @version     1
// @grant       GM_blobURI
// @grant       baseDomain
// @grant       isolateFetch
// ==/UserScript==

"use strict";

[...document.querySelectorAll("img")].forEach(img => {
  fetch(img.src).then(rsp => {
    return rsp.blob()
  }).then(blob => {
    let [gmuri, revoker] = GM_blobURI(blob)
    img.src = gmuri
  }).catch(err => {
    console.log(err)
  })
})
```

### DIFF
--- a/modules/sandbox.js
+++ b/modules/sandbox.js
@@ -12,6 +12,7 @@ Cu.import("chrome://greasemonkey-modules/content/storageFront.js");
 Cu.import("chrome://greasemonkey-modules/content/util.js");
 Cu.import("chrome://greasemonkey-modules/content/xmlhttprequester.js");
 Cu.import("chrome://greasemonkey-modules/content/extractMeta.js");
+Cu.import("chrome://greasemonkey-modules/content/scriptProtocol.js");
 
 var gStringBundle = Cc["@mozilla.org/intl/stringbundle;1"]
     .getService(Ci.nsIStringBundleService)
@@ -26,6 +27,16 @@ var gMaxJSVersion = "ECMAv5";
 
 
 function createSandbox(aScript, aContentWin, aUrl, aFrameScope) {
+  
+  let innerId = aContentWin.QueryInterface(Ci.nsIInterfaceRequestor).
+    getInterface(Ci.nsIDOMWindowUtils).currentInnerWindowID;
+  
+  // let the debuggers know that the sandbox is associated with a tab and addon
+  let metadata = {
+      "inner-window-id": innerId,
+      //addonId: "{e4a8a97b-f2ed-450b-b12d-ee082ba24781}"
+  };
+  
   if (GM_util.inArray(aScript.grants, 'none')) {
     // If there is an explicit none grant, use a plain unwrapped sandbox
     // with no other content.
@@ -36,6 +47,7 @@ function createSandbox(aScript, aContentWin, aUrl, aFrameScope) {
           'sandboxName': aScript.id,
           'sandboxPrototype': aContentWin,
           'wantXrays': false,
+          'metadata': metadata
         });
     // GM_info is always provided.
     injectGMInfo(aScript, contentSandbox, aContentWin);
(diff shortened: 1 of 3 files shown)


janekptacijarabaci:
A small comment:

1) Ad `loadInfo.externalContentPolicyType`
(https://github.com/greasemonkey/greasemonkey/pull/2420/commits/410b33cf93b306cf9a724c00d4289ad7219dfb37#diff-053bfe792f102bcaabbc0e694010e817R20)
Valid only for Firefox 44+, see also https://github.com/greasemonkey/greasemonkey/blob/3.8/modules/requestObserver.js#L27

2) How do you deal with incompatibility? (Greasemonkey/Tampermonkey...)

the8472:
> How do you deal with incompatibility?

Scripts can fallback via GM_XHR/the unprivileged XHR and `URL.createObjectURL()`. They simply won't work on pages with CSP, but they wouldn't work there without these features either.

In other words, you will only need these things if you're already having trouble with CSP.

the8472:
@arantius I'd like a little yay/nay whether you will accept something in this general direction. I'm open about the details.

arantius:
It's worth discussing, but my default position right now is I know that WebExtensions are coming and I'm not excited to add brand new features that are likely to break when that arrives.

the8472:
Webextensions [are already doing](http://searchfox.org/mozilla-central/rev/b92ae9a263a05efefc0110eaad5e56369c9f1b10/toolkit/components/extensions/ExtensionContent.jsm#391) some of this, and facing similar [CSP problems](https://bugzilla.mozilla.org/show_bug.cgi?id=1267027) beyond that.

the8472:
I mean we could also just skip the first two grants and change the behavior to always match what webextensions do. That would leave `GM_blobURI`, which could be specified to maybe or maybe not do any more than `URL.createObjectURL()`.

I just wanted to provide explicit opt-ins to avoid possibly breaking existing scripts.

the8472:
After some investigating it seems like we can skip the other things and limit it to adding `GM_blobURI`. 

Attempting to bypass CSP through separate XHR/fetch constructors does not set Referer and Origin headers properly. So that would probably break existing userscripts.

arantius:
Great, you're saying you think this feature *will* continue to be possible after a port of GM to webextensions?  If so it's worth proceeding with just `GM_blobUri` (capitalization?).

the8472:
> you're saying you think this feature will continue to be possible after a port of GM to webextensions? 

I derive a "yes" from the following observations:

- there are several CSP-related bugs filed against webext which they intend to fix
- fixing CSP also means getting blobs to work
- when they eventually fix it we can implement `GM_blobUri` as `URL.createObjectURL()` or a data: URI, which which is what script authors would have to use as fallback for feature-detecting `GM_blobUri` anyway.

I guess for maximum forward-compatibility I can return a promise instead of an URI. That way we would also have the option to chain blob -> filereader -> data URI or take a trip through the background page under webextensions.

In other words, future maintenance costs should be minimal.

mikhoul:
> In other words, future maintenance costs should be minimal.

:+1:

the8472:
Hrm, after some more testing it seems like i can make it work for CSPs that restrict to `self` and for types like images, but in some other cases it doesn't work.

So short of weakening CSP for a site I have no complete solution yet.

PR #2459, opened 2016-12-11
bump ff to 53, add seamonkey support

neb9:
can confirm it otherwise works with SM


janekptacijarabaci:
Ad SeaMonkey:
It isn't enough just to create `targetApplication`, more changes are needed.
See e.g. https://sourceforge.net/projects/gmport/

Croydon:
This seems now to be impossible due to the move to WebExtension, right?

Issue #2506, opened 2017-05-23
Component returned failure code: 0x8000ffff (NS_ERROR_UNEXPECTED) [nsISyncMessageSender.sendSyncMessage]


fdellwing:
I get the following error while running a userscript more than ~5 times at the same time. (Yes, this is a very likely situation for this script, because you open a lot of new tabs)

> 
> 19:48:41.905 Exception { message: "Component returned failure code: 0x…", result: 2147549183, name: "NS_ERROR_UNEXPECTED", filename: "chrome://greasemonkey-modules/conte…", lineNumber: 59, columnNumber: 0, data: null, stack: "GM_ScriptStorageFront.prototype.get…", location: XPCWrappedNative_NoHelper } 1 mm_skills_0_21.user.js:936:3
> 
> filename: chrome://greasemonkey-modules/content/storageFront.js
> message: Component returned failure code: 0x8000ffff (NS_ERROR_UNEXPECTED) [nsISyncMessageSender.sendSyncMessage]
[...]

Issue #2503, opened 2017-04-29
GM_getValue in AJAX callback sometimes throwing exception since Firefox 53


ccd0:
Seems to be a race condition of some sort. Only seems to happen when GM_setValue has been called recently.

Script to reproduce:
```
// ==UserScript==
// @name        GM_setValue test
// @namespace   test
// @include     http://example.com/
// @version     1
// @grant       GM_getValue
// @grant       GM_setValue
// ==/UserScript==

function getTest() {
  try {
    let n = GM_getValue('foo');
    console.log('get', n);
  } catch(err) {
    console.error(err);
  }
}

for (let i = 0; i < 10; i++) {
  let x = new XMLHttpRequest();
  x.open('GET', 'http://example.com/');
  x.onload = function() {
    console.log(x.status, x.statusText);
    getTest();
  };
  x.send();
}

let t = Date.now();
console.log('set', t);
GM_setValue('foo', t);
```

Tested with:
Greasemonkey	3.11
Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:53.0) Gecko/20100101 Firefox/53.0

Here's one of the errors thrown:
"Component returned failure code: 0x8000ffff (NS_ERROR_UNEXPECTED) [nsISyncMessageSender.sendSyncMessage]"

GM_ScriptStorageFront.prototype.getValue@chrome://greasemonkey-modules/content/storageFront.js:59:15
hitch/<@chrome://greasemonkey-modules/content/util/hitch.js:27:12
getTest@file:///C:/Users/anon/AppData/Roaming/Mozilla/Firefox/Profiles/xxxxxxxx.default/gm_scripts/GM_setValue_test/GM_setValue_test.user.js:12:13
x.onload@file:///C:/Users/anon/AppData/Roaming/Mozilla/Firefox/Profiles/xxxxxxxx.default/gm_scripts/GM_setValue_test/GM_setValue_test.user.js:24:5
waitForSyncCallback@resource://services-common/async.js:98:7
makeSpinningCallback/callback.wait@resource://services-common/async.js:168:27
promiseSpinningly@resource://services-common/async.js:234:12
get changedIDs@resource://services-sync/engines.js:98:5
_saveChangedID@resource://services-sync/engines.js:139:5
addChangedID@resource://services-sync/engines.js:160:7
notifyEvent@chrome://greasemonkey-modules/content/sync.js:229:11
[...]

PR #2507, opened 2017-05-24
GM_...Value - sometimes throws an errors: 0x8000ffff (NS_ERROR_UNEXPECTED) [nsISyncMessageSender.sendSyncMessage]

janekptacijarabaci:
Ad:
#2503, #2506

See also:
#2333

The suggestion.


makyen:
I can confirm that this change results in the exceptions not being thrown in Firefox 53.0.2 (tested by [disabling signature checking](https://stackoverflow.com/a/42403531/3773011)). Testing was performed with a script that was throwing exceptions 100% of the time in FF53-53.0.2 (I've not tested in 54+). I have not tested with FF running with multiprocess enabled.

I would not consider the limited use of this change which I have made to be anywhere near to complete testing.

Issue #2502, opened 2017-04-29
"Cannot request arbritrary files" error in version 3.11


gabrielfin:
There's been a recent change in PR #2486 which broke my scripts.
I'm using linux, and I have some scripts which have symlinks to a Dropbox folder. So basically: `gm_scripts/myscript` links to `~/Dropbox/myscript`

The recent update throws the error "Cannot request arbritrary files", because apparently it's trying to access `~/Dropbox/myscript/somerequiredfile.js` instead of `gm_scripts/myscript/somerequiredfile.js`

arantius:
This is not a supported configuration.

PR #2486, opened 2017-03-05
Fix e10s sandbox preventing scripts to load

RealDolos:
Fixes #2485

Turns out the basic `@resource` handling stuff was easy enough, with a caveat:
It's not possible to structured-clone some `responseType`s. Of course, this didn't work before already, but when trying to structured-clone some types via IPC, the child process will actually crash o_O. This was the case too for `blob`, thus I opted to load that to as `ArrayBuffer` and reconstruct the `Blob` again child-side.

Tested in e10s windows, but also in non-e10s windows to make sure stuff didn't break.
I *think* I didn't use features above the `minVersion`...

Issue #2485, opened 2017-03-04
e10s: user scripts fail to load due to the content process sandbox


RealDolos:
Firefox gradually introduces/introduced a sandbox for content processes[1]
Said sandbox will all block file accesses[2] when the level is configured to be at least 2, which is the case for Nightly at the moment. The sandbox works at least on macOS, meaning user scripts fail to load when e10s is enabled.

STR
===
- Get Nightly and install GM
- Create a new user script with whatever content (`console.log("hello, world");`)
- Load a page where this user script should run

Expected:
User script runs

Actual:
User script fails to run with the subscript loader bailing with
```text
Error opening input stream (invalid filename?): file:///Users/dolos/gmtest.p/gm_scripts/test/test.user.js
```

`Console.app` where sandbox violations are logged in macOS
```
SandboxViolation: plugin-container(35814) deny(1) file-read-metadata /Users/dolos/gmtest.p/gm_scripts/test/test.user.js
```

Workarounds
===
- Disable e10s
- Set `security.sandbox.content.level` to `1` or `0` lowering the sandbox protections

Remedies
===
Sync-message the main process and have it load the script source, then `evalInSandbox` the contents.

I prodded the source a bit till it *seemed* to work. Also `@resource` is not addressed at all but would work similarly I guess. Due to these caveats, no pull request.
```diff
diff --git a/components/greasemonkey.js b/components/greasemonkey.js
index 752a16c9..dccc9bc6 100644
--- a/components/greasemonkey.js
+++ b/components/greasemonkey.js
@@ -77,6 +77,10 @@ function startup(aService) {
       'greasemonkey:script-install', aService.scriptInstall.bind(aService));
   parentMessageManager.addMessageListener(
       'greasemonkey:url-is-temp-file', aService.urlIsTempFile.bind(aService));
+  parentMessageManager.addMessageListener(
+      'greasemonkey:getScript', m => {
+        return GM_util.fileXhr(m.data, "application/javascript,charset=utf-8");
+      });
 
   // Yes, we have to load the frame script once here in the parent scope. Why!?
   globalMessageManager.loadFrameScript(
diff --git a/modules/sandbox.js b/modules/sandbox.js
index 7c91f2c1..9284f2ee 100644
--- a/modules/sandbox.js
+++ b/modules/sandbox.js
@@ -182,12 +182,39 @@ function injectGMInfo(aScript, sandbox, aContentWin) {
   });
 }
 
+const remote = (function() {
[...]

Issue #1230, opened 2010-11-25
document.write() throws security error


arantius:
Firefox: `Mozilla/5.0 (Windows; U; Windows NT 6.1; en-US; rv:1.9.2.12) Gecko/20101026 Firefox/3.6.12`
Greasemonkey: master, commit ccdefe12f733a576710ea459be6612633c71caa8
Script: `document.write('working');`

Result: Error console shows `Error: Security error
Source File: file:///.../test.user.js
Line: 10`, script stops running.
Expected: Document content changes.

Also observed w/ GM 0.8.6.

janekptacijarabaci:
Firefox 40+:
Result: Browser console shows "The operation is insecure.", script stops running.

__However, document.write works in devtools Scratchpad.__

janekptacijarabaci:
RESOLVED FIXED for GM 3.5beta1+:
https://github.com/greasemonkey/greasemonkey/blob/3.5/modules/sandbox.js#L180

PR #2523, opened 2017-07-09
[suggestion] e10s: The content process sandbox (evalInSandbox -> loadSubScript)

janekptacijarabaci:
Add #2485, #2486

Reasons:
- The potential cache return - see: https://github.com/greasemonkey/greasemonkey/pull/2486#issuecomment-285957821
- Backward compatibility (partially - see #2502) - see: https://github.com/greasemonkey/greasemonkey/issues/2485#issuecomment-310886898 (#1230)

But I didn't try on Mac OS.


RealDolos:
This will create a ton of blobs and a ton of URLs that never get revoked, creating a memory leak. It will also lose the script URL information (when debugging)

janekptacijarabaci:
@RealDolos 
Thank you for your feedback.

> It will also lose the script URL information (when debugging).

See also #1844 and #2521 (AFAIK: I do not think it's a big problem).

I think backward compatibility should be maintained (as much as possible) - even in recent versions.
But it's just a suggestion.

the8472:
the blob URI should probably created once in the parent process.

* Open DOM `File` object
* turn into blob uri
* add it to the IPC script data.

Once a script is updated/unregistered the blob uri can be revoked in the parent process

Issue #2536, opened 2017-08-03
Broken in Nightly: nsIURI.path renamed to nsIURI.pathQueryRef


RealDolos:
Somehow the mozilla folks found it necessary to do this purely cosmetic change and break a ton of stuff in the process incl their own testing infrastructure (so no idea if that commit will actually stick in the long run).

https://bugzilla.mozilla.org/show_bug.cgi?id=1326520

Affected in greasemonkey are at least:
 - Update checking (because it tries to rename `user.js` to `meta.js`)
 - `MatchPattern` i.e. `@match`

jasontibbitts:
That PR does fix the issue for me.  It's so nice to have greasemonkey working again.

arantius:
Ah, this is unnecessary.

PR #2541, opened 2017-08-09
FF 57 (with XUL/XPCOM - which it still supports)

janekptacijarabaci:
## 1)

__FF 57: nsIURI.path renamed to nsIURI.pathQueryRef__
Ad #2536 (#2540)

## 2)

__FF 57: Remove nsIPrefBranchInternal__

Throws an error in Browser Console:
```
NS_ERROR_XPC_BAD_CONVERT_JS:
Could not convert JavaScript argument arg 0 [nsISupports.QueryInterface]
prefmanager.js:138
```

See:
https://bugzilla.mozilla.org/show_bug.cgi?id=1374847

### DIFF
--- a/modules/prefmanager.js
+++ b/modules/prefmanager.js
@@ -135,8 +135,15 @@ GM_PrefManager.prototype.watch = function(prefName, watcher) {
   // store the observer in case we need to remove it later
   this.observers.set(watcher, observer);
 
-  this.pref.QueryInterface(Components.interfaces.nsIPrefBranchInternal)
-      .addObserver(prefName, observer, false);
+  // http://bugzil.la/1374847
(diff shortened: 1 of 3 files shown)


janekptacijarabaci:
But see also:
https://blog.mozilla.org/addons/2017/08/10/upcoming-changes-compatibility/

> All legacy add-ons will have strict compatibility set, with a maximum version of 56.*. This is the end of the line for legacy add-on compatibility. They can still be installed on Nightly with some preference changes, but may break due to other changes happening in Firefox.

> Related to this, you won’t be able to upload legacy add-ons that have a maximum version set higher than 56.*.

@arantius
If you think you can close it...

PR #2593, opened 2017-09-28
package.sh: other providers for getting version

aeruder:
package.sh tries to extract the version from manifest.json.
Unfortunately the method used is specific to GNU sed, so the command
fails on OS X and other operating systems with the BSD userland.

Add in a few other options for one-liners for just grabbing the version
out:

  jq: the best as it can actually just parse the json
  gsed: if this is available, it should for sure be GNU sed, often
        BSDs will install GNU sed under this name.
  perl: Available almost everywhere and will also do the trick

And last, if nothing else is available, fall back on the original
sed command which usually only works on linux and certain windows
environments.

### DIFF
--- a/package.sh
+++ b/package.sh
@@ -3,7 +3,18 @@
 set -e
 set -x
 
-VER=$(sed -n -e '/"version"/{ s/.*: "//; s/".*//; p; q}' manifest.json)
+function get_version {
+  while read line ; do
+    if [[ "$line" == *\"version\"* ]]; then
+      line="${line##*: \"}"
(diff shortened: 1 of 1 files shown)


arantius:
Can you find a single portable way to do this?

Sxderp:
Or a separate oneliner, using pipes:
`VER=$(grep '"version"' manifest.json | sed 's/"version": "\(.*\)",/\1/' | tr -d ' ')`
Works on my FreeBSD VM (so should work on Mac OSX) and Cygwin (which I believe is GNU based).

PR #2595, opened 2017-09-28
package.sh: use bash directly for getting version

aeruder:
package.sh tries to extract the version from manifest.json.
Unfortunately the method used is specific to GNU sed, so the command
fails on OS X and other operating systems with the BSD userland.

Since we're already depending on bash, just use bash directly to extract
the version.  It isn't pretty but avoids external dependencies (besides
zip, of course).

### DIFF
--- a/package.sh
+++ b/package.sh
@@ -3,7 +3,18 @@
 set -e
 set -x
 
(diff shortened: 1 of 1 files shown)


arantius:
f68ccc0e579127e659dd95477f14654434cf1984 obsoletes this

Issue #2602, opened 2017-09-30
Upgrade from 3.15 to 3.16 makes script very slow


LoveIsGrief:
Hello,

I have a [custom script](https://gist.github.com/anonymous/f463ba1f2e559ba9b8d95862cf768681) that helps me on reddit and in version 3.15 it runs barely using any resources, but in 3.16 it suddenly slows the whole page down to 0.5 fps.

[This seems to be the only change](https://github.com/greasemonkey/greasemonkey/compare/3.15...3.16) between the versions, but it significantly slowed down the script.

```
$ firefox --version
Mozilla Firefox 55.0.2

$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 16.04.3 LTS
Release:        16.04
Codename:       xenial

$ uname -a
Linux loneliverse 4.4.0-97-generic #120-Ubuntu SMP Tue Sep 19 17:28:18 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
```

ccd0:
Based on sticking a `dump` statement after the `this._convertScriptToWebext(script);` line, the code in
https://github.com/greasemonkey/greasemonkey/compare/3.15...3.16
seems to be running after every call to GM_setValue, which I don't think was intended since there's already a line here
https://github.com/greasemonkey/greasemonkey/blob/3.16/modules/storageBack.js#L84
that migrates just the one setting being changed.

Sxderp:
@LoveIsGrief, you think you could check out my [branch](https://github.com/Sxderp/greasemonkey/tree/prevent-double-migrate-value) and see if that fixes it for you. Also, thanks to @ccd0 for pointing this issue out.

LoveIsGrief:
@Sxderp I haven't migrated my script to 4.0 yet ☹️  Can't test it.

fdellwing:
Can confirm, that something in 3.14+ is making the getter and setter calls really slow.

arantius:
Please try, and let me know if it helps:

https://addons.mozilla.org/firefox/downloads/file/735451/greasemonkey-3.17beta1-fx.xpi

ccd0:
It fixes the slowdown I was seeing in my script; thanks.

fdellwing:
Works.

PR #2603, opened 2017-09-30
Prevents double migrate value #2602

Sxderp:
This should help alleviate #2602


Sxderp:
Fixed with 8fa55b33e3640e5d2224f250276a2d49856798da, closing.

Issue #2558, opened 2017-09-13
Improve script uninstall


arantius:
In at least two ways:

1. Provide an "undo" button.
2. Add it to the monkey menu.

Sxderp:
I think 2. should be merged into #2585.

Issue #2585, opened 2017-09-26
Refine the monkey menu script list


arantius:
Right now the monkey menu always lists every installed script every time.

If we keep this behavior, it should be split into "active on this page" and "the rest" sections, to make the first more relevant group more discoverable.

I've been toying with getting rid of the full-tab "manage" dialog in favor of doing all script management in the monkey menu (expanding the per-script display and options), but have not made a firm decision about that.  If that decision is positive, do as above.  If we keep the manage dialog, then perhaps we go back to the pre-4.0 style of only listing the active on this tab scripts.

Sxderp:
The problem I see is large lists of scripts. When drop-downs/button-menus get long it can be really inconvenient dealing with them.

Sxderp:
Another thing to consider are the people who don't like having extra buttons in the toolbar area. 

Aside from that I think, after mulling it over, I think doing script management through the menu could work out. There are a couple things that I think should have some consideration. 

1. Instead of listing scripts for the current page in the standard button menu, create a [page action](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/pageAction) button and list them there. I think this fits very nicely in what a page action should be. 
2. For the full list of scripts in the Monkey Menu, separate them out by namespace submenus. That way large lists of scripts should be manageable.

arantius:
> Another thing to consider are the people who don't like having extra buttons in the toolbar area.

You can customize this like button like any other part of the UI, putting it into the overflow or hamburger menu.  Browser Extensions get almost no UI besides the "browser action" which is a toolbar button.

> create a page action button and list them there

This doesn't feel right to me.  This would contain things only vaguely connected to the page.  Plus you can't customize page actions out of your toolbar ;-)  For some/many users this page action would ~always be visible.  On mobile, page actions have to fit into a very small space.  And what happens if two different extensions try to register a page action for the same page?

> The full list of scripts in the Money Menu should be separated by namespace submenus

I disagree quite strongly.  Namespace is an invisible property that only script authors understand; explaining to a normal user why a script is in a group would be unnecessarily difficult.  This is something that you're generally not in control of as a user, scripts would appear to be in a random order more than anything else.

I think I prefer the legacy style of "active" scripts (ran on this tab) in one group, while "inactive" (did not run on this tab) ones are separate.  Usually you're looking for something "running", no?

Sxderp:
[...]

PR #2610, opened 2017-10-06
Undo remove script for web manager

Sxderp:
Added an undo button in the web interface for managing scripts. Addresses 1. of #2558.

### DIFF
--- a/src/content/manage-user-scripts.css
+++ b/src/content/manage-user-scripts.css
@@ -45,3 +45,26 @@ li.user-script button {
   margin-left: 1ex;
   width: 100%; /* Shared within flex container. */
 }
+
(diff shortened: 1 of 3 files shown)


jerone:
This won't the userscript when you click the remove button and then immediately close your browser, right?

Sxderp:
> This won't [remove] the userscript when you click the remove button and then immediately close your browser, right?

Probably correct.
But since @arantius is inclined toward #2585 I don't particularly feel the necessity to find a remediation to the problem. If this pull request is accepted and therefore the management page sticks around I'll look into a fix.

Sxderp:
Closing due to recent commit removing the manage user script page.

Issue #2560, opened 2017-09-13
Switch back to `browser` API with promises


arantius:
https://github.com/mozilla/webextension-polyfill

Consider switching back to the `browser`/promise based WebExt API, with a polyfill for other browsers if/when support is covered, there.

Sxderp:
I think this was an accidental close. Commit 22fa2cd should have been #2650

arantius:
Yes, thank you.

PR #2614, opened 2017-10-12
[WIP] Switch to `browser` API with Promise

yfdyh000:
solve #2560. This is a WIP, it needs more work and testing.
1.  `sendResponse` callback to promise, I'm not know how to do it. This looks blocks running in Chrome.
2. There is not a full-featured (APIs) userscript and a test process to make sure it works properly, and even an automated tests, code coverage.


arantius:
First, huge thanks for your time putting this together.

That said, it's reasonable to reach out before making large changes like this.  I had existing plans for how to arrange which work got done when.  This was low priority because it doesn't help get us to a functional state.  And we've already got merge conflicts for your pull request.

As you point out, there's no unit tests (yet).  Putting those in is one of the things I was planning to do before a big refactor like this.

arantius:
There's since been significant drift, and I'm more confident that this will break tests.  As/when this change is made, it should be very incremental, in small parts.

Issue #2639, opened 2017-10-31
Improve accessibility


arantius:
I've received an email directly from a vision impaired user.  Accessibility of both the monkey menu and the script editor are currently poor and should be improved.

Sxderp:
Like, screen reader accessibility? Could you provide more details?

arantius:
Yes.  I don't have a lot more detail (yet), but I also don't want the task to be forgotten -- therefore this issue.

Best I can say for now:  Try closing your eyes and using GM 4.0.  How does it go?  Poorly I bet.  I'm guessing we need, at a minimum, some aria attributes all over the monkey menu.

Sxderp:
Keyboard navigation may also be a good idea. At work there's someone I service that I know uses scripts / keyboard to do things because they don't have the dexterity to position the mouse. Currently I'm unable to tab into (after the menu is already opened)[1] nor navigate with arrow keys.

[1] As for opening the monkey menu, I don't think WebExtensions can modify that behavior at all. I could be missing an API though.

arantius:
Yes, better keyboard navigation helps everybody.  Right now we're all <div> with onclick, nothing that tab would focus.  Perhaps we should be more <a> or similar?

I'm trying to/mostly like working "like a menu" so up/down arrows would make sense but I wonder how easy/hard that will be.

arantius:
https://bugzilla.mozilla.org/show_bug.cgi?id=1324255

Sxderp:
> Perhaps we should be more <a> or similar?
This would probably be better. I'm pretty sure the items would be highlight-able then. I recently created a Javascript gallery / pager in which you could switch between a gallery view or a photo paging view. I used <a> tags and anchors along with [onhashchange](https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onhashchange) in order to navigate between views and photo paging. Assuming that's still available in the popup it might be a viable option.

arantius:
The above (merged) PR improves things, but needs work.

jsl303:
Thank you for all the efforts to make Greasemonkey accessible.
For Windows 10 (starting 1703 Creators Update), you could use built-in screen reader called Narrator.
https://support.microsoft.com/en-us/help/22798/windows-10-narrator-get-started
For earlier Windows, you could test the accessibility using a free screen reader called nvda.
https://www.nvaccess.org/files/nvda/documentation/userGuide.html
[...]

PR #2660, opened 2017-11-11
Initial Accessibility Improvements for Monkey Menu

Sxderp:
Partial implementation for #2639.
Basic keyboard navigation is supported. Up/down arrows.

### DIFF
--- a/src/browser/monkey-menu.html
+++ b/src/browser/monkey-menu.html
@@ -61,19 +61,19 @@
 
   <hr>
 
-  <a href="#open-tab_https://www.greasespot.net/"
+  <a href="#https://www.greasespot.net/"
       class="menu-item">
     <div class="icon fa fa-fw fa-link"></div>
     <div class="text">Greasemonkey home page</div>
   </a>
 
-  <a href="#open-tab_https://wiki.greasespot.net/"
+  <a href="#https://wiki.greasespot.net/"
       class="menu-item">
     <div class="icon fa fa-fw fa-link"></div>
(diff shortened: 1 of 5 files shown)


src/browser/monkey-menu.html
```diff
@@ -40,66 +41,82 @@
...
+  <a href="#open-tab_https://wiki.greasespot.net/"
+      class="menu-item">
     <div class="icon fa fa-fw fa-link"></div>
     <div class="text">Greasemonkey Wiki</div>
-  </div>
+  </a>
 
-  <div class="menu-item"
-      data-url="https://wiki.greasespot.net/User_Script_Hosting">
+  <a href="#open-tab_https://wiki.greasespot.net/User_Script_Hosting"
```
arantius: I'm not excited about making up custom string-concatenated formatting of various values that have to be carefully parsed back out again.  I think there's a better way to do this.  Especially in this case, we need to have a JS handler set to avoid navigating to this fake anchor anyway.  So just put the real URL, cancel the event, and do the create-a-tab API call with the actual HREF value.

Sxderp:
Updated with the suggestion, and a little bit of css / visual formatting.

Normally I'd rebase onto master but I wasn't sure how that would effect the review comment, so I just merged.

Sxderp:
And somehow I've got a garbage commit in this. Screw it. I'm just going to rebase on master.

Sxderp:
Ouch, I hadn't been keeping updated with this particular branch. Seems like it was failing tests. Might want to pullout / revert until those can be resolved.

Issue #2670, opened 2017-11-15
GM 4 Scripts Error: expected expression, got ')'


erosman:
Hi

Working scripts under GM 3 fail under GM 4

```
Could not execute user script: [Greasemonkey Script .../....; 1.0] 
 Error: expected expression, got ')'
Stack trace:
Async*@moz-extension://ffbb8d40-70fc-42b4-9d63-f5ac882801f8/src/bg/execute.js:23:5
  execute.js:29:7
```

Any idea what the cause is?

**Note:** There is no JS error in the code. (tested)

After testing .... it seem the following fails

```
// ==UserScript==
// @name          Clean URL
// @namespace     erosman
// @description   Clean URLs
// @include       *
// @grant         none
// @author        erosman
// @version       1.0
// ==/UserScript==


(function() { 


})(); // end of anonymous function
```

Fix: adding an extra line or removing the end comment eliminates the error.

The following works fine .... (added a blank line)

```
// ==UserScript==
// @name          Clean URL
// @namespace     erosman
// @description   Clean URLs
// @include       *
// @grant         none
// @author        erosman
// @version       1.0
// ==/UserScript==
*/

(function() { 


})(); // end of anonymous function

```

Following works as well ... (removed the end comment)

```
// ==UserScript==
// @name          Clean URL
// @namespace     erosman
// @description   Clean URLs
// @include       *
// @grant         none
// @author        erosman
// @version       1.0
// ==/UserScript==
*/

(function() { 


})();
```

Sxderp:
The way the parsing works is to create a function for the userscript and then call it. The current implementation defines the script on a [single line](https://github.com/greasemonkey/greasemonkey/blob/71e9be5e38f984977f128bb439130ada6446c185/src/user-script-obj.js#L252). If the userscript ends in a comment than it causes the function definition to be invalid.

Sxderp:
Submitted a pull to resolve the issue.

arantius:
Short term workaround: edit your scripts so they don't have a `//` line comment on the last line.

Will be fixed in next release.

erosman:
> Short term workaround: edit your scripts so they don't have a // line comment on the last line.

Obviously as pointed out in my initial post, I had already worked that out. ;)

arantius:
Sorry you edited that in after your first post, so I didn't see it (read the message as delivered to my email).

erosman:
[...]

PR #2674, opened 2017-11-15
Fix failing with trailing comments in scripts

Sxderp:
See #2670 

When a user script had a trailing comment it would comment out part
of the eval string. Catch them so they don't interfere.

### DIFF
--- a/src/user-script-obj.js
+++ b/src/user-script-obj.js
@@ -249,10 +249,7 @@ window.EditableUserScript = class EditableUserScript
     // of the generated parts.
     this._evalContent
         // Note intentional lack of line breaks before the script content.
-        = `try {
-        (function scopeWrapper(){
-        function userScript() {
-        ${this._content}
+        = `try { (function scopeWrapper(){ function userScript() { ${this._content}
         /* catch open comment blocks */ }
         const unsafeWindow = window.wrappedJSObject;
         ${this.calculateGmInfo()}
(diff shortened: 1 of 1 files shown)


src/user-script-obj.js
```diff
@@ -249,7 +249,11 @@ window.EditableUserScript = class EditableUserScript
     // of the generated parts.
     this._evalContent
         // Note intentional lack of line breaks before the script content.
-        = `try { (function scopeWrapper(){ function userScript() { ${this._content} }
+        = `try {
+        (function scopeWrapper(){
+        function userScript() {
+        ${this._content}
+        /* catch open comment blocks */ }
```
arantius: A) I prefer full/grammatical sentences in comments
B) Something like `/* Force a line break after `_content` in case its final line contains a line comment. /*` seems much more useful; just "catch open comment blocks" is so terse as to be hard to understand later.

src/user-script-obj.js
```diff
@@ -249,7 +249,11 @@ window.EditableUserScript = class EditableUserScript
     // of the generated parts.
     this._evalContent
         // Note intentional lack of line breaks before the script content.
-        = `try { (function scopeWrapper(){ function userScript() { ${this._content} }
+        = `try {
```
arantius: Please see the comment on the line above.

Adding a line after `_content` is good to fix this bug.  Adding lines above it makes the lines in reported errors harder to use.

Sxderp: Ah, missed that. I'll submit a commit the removes the lines.

src/user-script-obj.js
Sxderp: Alright. Will adjust.

arantius:
This fix is in Version 4.1beta4.  Testing to confirm the fix is appreciated!

https://addons.mozilla.org/en-US/firefox/addon/greasemonkey/versions/beta?page=1#version-4.1beta4

Issue #2690, opened 2017-11-17
"Save script" shortcut should match platform (ctrl vs. cmd)


adaugherity:
After upgrading to FF 57 (which upgraded GM to v4), my script no longer works.  I tried to fix it via the editor, but appeared to be unable to save my changes as there is no save button and Cmd-s brought up the Firefox Save Page As... dialog.

I eventually discovered https://groups.google.com/forum/#!topic/greasemonkey-users/GLTDE-G9LZI indicating there are not (yet) any buttons, and Ctrl-S is the shortcut to save.  That did indeed work, but is unexpected on macOS.  Can we please make the shortcut be ⌘S (Cmd-s) on Mac?

arantius:
This is in build 4.1beta5 ( https://addons.mozilla.org/firefox/downloads/file/793791/greasemonkey-4.1beta5-an+fx.xpi?src=devhub ).  Testing and confirming would be appreciated!

adaugherity:
Tested 4.1b5 on Mac, and the save shortcut is now cmd-s. 👍 

I do see some warnings logged to the browser console -- are these known issues?

Upon opening the editor: 
```
Loading failed for the <script> with source “moz-extension://e8447a6c-bfa1-8e46-a717-cd76d4efe15c/third-party/codemirror/addon/scroll/annotatesscrollbar.js”.
edit-user-script.html:16
```
Saving a script:
```
updater has no added remotes to handle
user-script-obj.js:349:9
```

Sxderp:
> Upon opening the editor:

Not sure about that. Though I think it may be a bug, as far as I know all the codemirror things should load properly.

> Saving a script:

Standard. Just some debug info detailing that there are no `@require` or `@resource` tags.

xor10:
> Not sure about that. Though I think it may be a bug, as far as I know all the codemirror things should load properly.

Typo.
```diff
diff --git a/src/content/edit-user-script.html b/src/content/edit-user-script.html
index 9c1d1b8..d94e7e2 100644
--- a/src/content/edit-user-script.html
+++ b/src/content/edit-user-script.html
@@ -13,7 +13,7 @@
 <script src="../third-party/codemirror/addon/dialog/dialog.js"></script>
 <link rel="stylesheet" href="../third-party/codemirror/addon/dialog/dialog.css">
 
-<script src="../third-party/codemirror/addon/scroll/annotatesscrollbar.js"></script>
[...]

PR #2691, opened 2017-11-17
Issue #2690 Define a codemirror "save" command

badbrainz:
CodeMirror's default keymap already handles 'ctrl-s' and even runs
a 'save' command when available.

see docs
https://codemirror.net/doc/manual.html#commands

### DIFF
--- a/src/content/edit-user-script.js
+++ b/src/content/edit-user-script.js
@@ -6,11 +6,10 @@ var editor = CodeMirror(
     // TODO: Make appropriate options user-configurable.
     {
       'tabSize': 2,
-      'extraKeys': {
-        'Ctrl-S': onSave,
-      },
       'lineNumbers': true,
     });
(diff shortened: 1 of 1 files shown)


Sxderp:
Does this also deal with #2690?

badbrainz:
Yeah. CodeMirror initially sets its default keymap to pcDefault or macDefault [#L47](https://github.com/codemirror/CodeMirror/blob/b881f2520461c7fc98ca1673071d0eb74ddd7c39/src/input/keymap.js#L47). If you need the cmd/ctrl prefix check what default is:
```
let ctrl = CodeMirror.keyMap.default === CodeMirror.keyMap.macDefault ? 'Cmd' : 'Ctrl';
...
'extraKeys': {
  [ctrl + '-S']: onSave,
}
```

Issue #2698, opened 2017-11-18
Change the icon of "Greasemonkey is disabled" to something more meaningful


tiansh:
An [O Mark](https://en.wikipedia.org/wiki/O_mark) is used currently, which is confusing and just telling me "correct" or maybe "enabled". [No Symbol](https://en.wikipedia.org/wiki/No_symbol) or [X Mark](https://en.wikipedia.org/wiki/X_mark) may be a better choice.

Version info: Greasemonkey 4.0

Sxderp:
Agreed

arantius:
It's just a (round) checkbox that can either be checked or not.  Why isn't that meaningful?

I'd interpret the X mark in the opposite way: that's a checked checkbox, that means enabled to me.

tobbexiv:
To me the o usually means a bullet point list or "check here". Additionally this does not change, only the content changes. Thus, I think its clear especially as there is a text next to it so if you really do not understand it, you can read the text.

arantius:
(only) a check mark when enabled and a complete empty space when disabled could also be an unambiguous improvement perhaps?

tiansh:
tick vs. none may be a good choice, just like what firefox does.

![tick](https://user-images.githubusercontent.com/6500326/33000043-a05f8ef2-cde0-11e7-8058-fc51a3500aa3.png)

S-ed:
@tiansh how people suppose to know there should be a 'tick' if there's nothing?

Sxderp:
> how people suppose to know there should be a 'tick' if there's nothing?

Considering Mozilla uses it in their UI I assume there's at least _some_ thought behind it. Although, may be garbage thoughts considering what I've seen from UI/UX guys.

S-ed:
Yeah, the tick is only working for "Toggle Tools" and "Developer Toolbar"...

arantius:
If you think this needs to change: Please open http://fontawesome.io/icons/ and suggest by name the *two* icons that you think should represent A) the enabled and B) the disabled status.  And why the selection of these two is better than the status quo.

tiansh:
here are some choices, imo

enabled | disabled | reason
---- | ---- | ----
toggle-on | toggle-off | widely known, and commonly used for enable / disable features (prefered)
check-circle-o | ban | minimal modification and no ambiguous
check-square-o | square-o | an checkbox input normally use square, at least on windows and mac
[...]

PR #2699, opened 2017-11-18
Change GM disabled symbol to X-mark

Sxderp:
#2698

https://imgur.com/a/OICNO


arantius:
Definitely do not like the circle-with-check turning into an X.  Both states should make sense, and the transition between them.

Sxderp:
Closing as discussion continues on the other thread.

Issue #2684, opened 2017-11-16
The name of unknown key in userscript meta data warning is truncated: @supportURL triggers an 'Expected "author", "description", ... or "version" but "s" found.' warning


juozaspo:
Having `@supportURL` or any other unknown key such as gibberish in userscript meta info triggers CodeMirror lint warning with truncated key name: 
`Expected "author", "description", "downloadURL", "exclude", "grant", "homepageURL", "icon", "include", "installURL", "match", "name", "namespace", "noframes", "require", "resource", "run-at", "updateURL" or "version" but "s" found.` 
Full text captured using firefox debug tools, screenshots below.
![pazymejimas 2017-11-16 15 14 43](https://user-images.githubusercontent.com/6626242/32893396-f636f4ca-cae1-11e7-947f-2874205f58f7.png)
![pazymejimas 2017-11-18 16 17 46](https://user-images.githubusercontent.com/6626242/32981217-30143fd6-cc7c-11e7-8d83-7e242c9c6ed7.png)
Greasemonkey 4.1beta3 and 4.1beta4

Sxderp:
Would be nice if it reported the entire string "supportURL."

Eselce:
You guessed it: Same with @copyright and "c" found... ;-)

arantius:
#2703 has been packaged in a new beta version:
https://addons.mozilla.org/firefox/downloads/file/830369/greasemonkey-4.2beta1-an+fx.xpi?src=devhub

Testing and confirmation would be appreciated!

Eselce:
You've probably already registered, that it works fine.

For me, it was @copyright (I use it, although GM doesn't accept it).

GM 4.1 (and earlier) mentioned "c", while this fix extends it to "copyright".
The whole word is underlined to mark the failure.

PR #2703, opened 2017-11-19
Issue #2684 Replace unhelpful lint message

badbrainz:
Swap the truncated part with the word that's in the editor doc.
Since we're computing the 'to' position (metaline parser doesn't
give us the one we want) change that as well.


src/content/edit-user-script.js
```diff
@@ -147,7 +147,17 @@ function onSave() {
 editor.on('swapDoc', doc => {
   if (doc.getMode().name == 'javascript') {
     doc.setOption('gutters', ['CodeMirror-lint-markers']);
-    doc.setOption('lint', true);
+    doc.setOption('lint', {
+      formatAnnotation(ann) {
+        if (ann.from.ch == 4) {
```
arantius: Why 4?

Is there a named constant available that would make this code more readable?

src/content/edit-user-script.js
```diff
@@ -147,7 +147,17 @@ function onSave() {
...
+          let to = doc.findWordAt(ann.from).head;
+          let invalidWord = doc.getRange(ann.from, to);
+          ann.message = ann.message.replace(/(.+")([^"]+)(".+)$/, `$1${invalidWord}$3`);
```
arantius: 80 columns max; please adopt formatting like the rest of the code base

src/content/edit-user-script.js
Sxderp: I'm not 100%, but I think the 4 is the column number of the `@` sign used for keywords.

badbrainz: It's the col number. I moved this to the lint function since it's meant for metadata.

src/content/edit-user-script.js
badbrainz: Noted.

arantius:

Issue #2705, opened 2017-11-19
Should scripts be re-running when going backwards and forwards though history?


vzjrz:
I don't know if this is a greasemonkey or firefox bug but after the new updates I noticed some unusual behavior from my scripts.
Scripts are running again on pages where they already ran when using the back and forward arrows. I checked with the old versions of gm and firefox and that's not how they used to work. In the old version of firefox I guess pages would completely rerender and therefore when scripts reran they would run on unmodified pages.
```
// ==UserScript==
// @name        Test
// @namespace   test
// @include     *://www.google.com/*
// @require     https://code.jquery.com/jquery-3.1.1.min.js
// @version     1
// @grant       none
// ==/UserScript==

(function () {
  'use strict';

  let tests = $('#test');
  
  // Needed with new updates:
  if (tests.length === 0) {
    console.log('Adding listener');
    document.body.addEventListener('click', e => alert('clicked!'))
  } else {
    console.log('Test div detected');
  }
  
  $('body').append('<div id="test">');
})();
```

1. Install this script and go to https://www.google.com/search?q=test
2. Click on a link to go to a different url
3. Go back
4. Go forward
5. Go back

In gm 3.17 this script would never print `Test div detected` but in gm 4.0 going forwards and back multiple times just keeps adding more and more divs.
Is this an intended change?

Sxderp:
Not able to reproduce with FF 59.0a1 + Latest master.

vzjrz:
Man I should really test things out before I write bug reports. I found the bug on another site and tried to write simple case for google.com but checking that out now it doesn't appear to work there. I have updated the steps to reproduce it.

Sxderp:
Cool. Was able to reproduce. I believe I know a fix. Will submit shortly.

arantius:
This is probably complicated by [fastback](https://wiki.mozilla.org/DocShell/Fastback), which if it applies then the answer is no, but if it does not then the answer is yes.

Sxderp:
In response to the comment on the pull (I think it's better to respond here then a commit comment).
I see your point, and it can somewhat be forcibly reproduced. The initial post in this issue had used google as a test subject. And for whatever reason google searches don't work with fastback (thus the original problem never appeared).

With the change performing a back / forward on google does not see the script executed. No `div` was added to the DOM.

Sxderp:
I had a thought and just tested with #2635 and it looks like that also solves the forward/backward button issue.

[...]

PR #2708, opened 2017-11-19
Do not re-execute code on forward / backward nav

Sxderp:
#2705

### DIFF
--- a/src/bg/execute.js
+++ b/src/bg/execute.js
@@ -6,7 +6,10 @@ TODO: Make document_start execution time work as intended.
 */
 
 function executeUserscriptOnNavigation(detail) {
-  if (false === getGlobalEnabled()) return;
+  if (false === getGlobalEnabled()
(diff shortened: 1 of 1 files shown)


src/bg/execute.js
```diff
@@ -6,7 +6,10 @@ TODO: Make document_start execution time work as intended.
 */
 
 function executeUserscriptOnNavigation(detail) {
-  if (false === getGlobalEnabled()) return;
+  if (false === getGlobalEnabled()
+      || detail.transitionQualifiers.indexOf('forward_back') !== -1) {
```
arantius: See comment on issue, it's possible for the entire after-scripts-run document to be cached, and it's possible not to.  What if you go forward ten times and then back ten times, what happens?

Sxderp:
Closing as this doesn't actually solve the issue. Unfortunately the details object doesn't have any indication on whether Firefox is navigating to a cached page or not.

Issue #2631, opened 2017-10-25
GM 4.0 does not install scripts from GitHub (CSP/sandbox issue)


Incurious:
GM doesn't install scripts from raw.github.com for some reasons

https://github.com/reek/anti-adblock-killer#anti-adblock-killer--reek
Here on "Step 3: UserScript" you can test 4 mirrors of one script. only github mirror doesn't open install dialog.

badbrainz:
There might be an issue with host permissions in this case. Content script aren't injected into top frames that are "about:blank" according to the moz docs.

xor10:
I think it's a Firefox CSP issue. Apparently, you cannot inject content scripts into pages with CSP sandbox directive. Turn off the CSP (set "security.csp.enable" to false) and try to install it. Do not forget to activate the CSP after installing.

![csp](https://cdn.ghacks.net/wp-content/uploads/2013/06/security-csp.png)

arantius:
https://bugzilla.mozilla.org/show_bug.cgi?id=1267027#c28

Sxderp:
Hm, currently at work and only have access to Firefox 52 ESR. Cannot reproduce, `security.csp.enable` is set to true, using currently released alpha 4.0 available on [AMO](https://addons.mozilla.org/en-US/firefox/addon/greasemonkey/?src=userprofile#beta-channel). What version of Firefox does this fail on? Might be a regression that should be brought up to the Mozilla folks. CSP should not affect content scripts.

arantius:
Happens to me on 56 (stable) and 57 (nightly).

arantius:
Also reproduces: https://gist.github.com/arantius/f6fd80b1efad368a45ca35567bc31b18 (click "raw")

54 works

arantius:
55 also works, 56+ is broken

https://bugzilla.mozilla.org/show_bug.cgi?id=1411641

adaugherity:
User script installation from GitHub *does* work in ViolentMonkey (2.8.18) in Firefox 57.  What are they doing differently?

Sxderp:
Not really, no. Violentmonkey directly injects scripts by creating `<script>` objects. Greasemonkey uses `tabs.executeScript()`. Both methods have their own nuances and issues.

---
[...]

Issue #2683, opened 2017-11-16
Editing user script causes @require to fail loading


vzjrz:
Firefox 57 GM 4.1beta3
I've upgraded from the older greasemonkey and now am in the process of fixing my scripts but I noticed that if you edit something and then save greasemonkey sometimes doesn't get the @require scripts.
```
// ==UserScript==
// @name          Test
// @include       *://github.com/*
// @require       https://code.jquery.com/jquery-3.1.1.min.js
// @require       https://greasemonkey.github.io/gm4-polyfill/gm4-polyfill.js
// @grant         GM.addStyle
// @grant         GM_addStyle
// @run-at        document-idle
// @noframes
// ==/UserScript==

(function () {
	'use strict';

	console.log($);
	console.log(GM_addStyle);
})();
```

```
 Script error:  
ReferenceError: $ is not defined
```
In order to get it to work I need to:

1. Do my edits
2. Save
3. Add a `?` character to the end of each @require url
4. Save
5. Remove the extra characters from the @require urls
6. Save

That seems to fix it.

StigNygaard:
I haven't investigated so much as vzjz, but I have had similar problems.

Eselce:
I've had several issues: No update of depending script resources after editing.

Sxderp:
I've actually run into this once before too. I just figured it was something in my environment due to screwy testing.

Can someone confirm if (from following the instructions in first post)

7. Make some changes
8. It happens again?

Sxderp:
#2689 may help with this.

StigNygaard:
Hi @Sxderp , how "deep" does fix for #2689 go? The reason I ask is, that as far as I remember when seeing the issue, it was not only an UI problem. The userscript also stopped working because it couldn't find the require'd sub-script.
Sorry, haven't had time to investigate more. So it's only as I remember it...

Pantalaim0n:
@Sxderp I can confirm the issue exists when making changes to the script, the @require does not work.

Sxderp:
Alright, I was able to reproduce this. Not sure if the execution workflow what I forced is similar to others.

1. Make some basic script (that works).
2. **Add** a @requires
3. Force an indexedDB error[1]. Do this by having two scripts with the same name. That is, rename the script you're working on to something already existing.
4. Save
5. Fix the error.
6. Save
7. Continue editing without touching the @requires
8. Save
9. Run the script and the @requires is not there. Further editing of the script does not solve the problem.

[...]

Issue #2643, opened 2017-11-03
Don't trigger install while globally disabled


arantius:
In legacy Greasemonkey: while globally disabled we would ignore navigation to scripts and not pop the install dialog.

In 4.0, navigation to an appropriate URL always pops the dialog.  We should restore the old behavior.

Issue #2641, opened 2017-11-02
Install via pop up window, not tab


arantius:
This issue exists primarily to track my desire to make this change, but inability to:

https://bugzilla.mozilla.org/show_bug.cgi?id=1402110

Opening dialogs is buggy.  I've written all the code but it doesn't work thanks to the linked Firefox bug.

Sxderp:
Of interest. In one of the [test branches](https://github.com/Sxderp/greasemonkey/tree/use-window-to-install-from-cache)[1] I've been working on I implemented a popup box. And the content painted correctly, without having to click on it[2]. Though the title prefix is still an issue. While the prefix is added but the `mox-extension://blahblahblah` is still there.

[1] The branch is one that I had setup detecting UserScripts in the background and then caching the content so that the XHR wasn't required to install new scripts. I don't think those changes particularly effect this issue though.
[2] On 59.0a1

arantius:
I've also got a branch for this ( https://github.com/arantius/greasemonkey/tree/install-dialog ) though I haven't tested in a while.

Sxderp:
Turns out the problem seems to be with `extensions.webextensions.remote`.

If that preference is set to `false` (default for Linux) then the window will be white and buggy. If the preference is set to `true` (default for Windows) then the window displays without problem.

arantius:
I've updated my branch, linked above, with a workaround.  I'd appreciate any comments on how this works.  It's a little bit janky, but maybe good enough?

Sxderp:
Android can be accommodated with something like (tested on my phone).

```javascript
chrome.runtime.getPlatformInfo(platform => {
  let url = chrome.runtime.getURL('src/content/install-dialog.html')
      + '?' + escape(JSON.stringify(message.userScript));

  if ('android' === platform.os) {
    chrome.tabs.create({'active': true, 'url': url});
  } else {
    ...
  }
});
```

---

Why `history.back()`? I'm in favor of just leaving the page open. That way the user can examine the script if they want before installing.

Sxderp:
> (tested on my phone)

Apparently I didn't test enough. Opening the install thing works on the phone, but install the script just fails. At the moment I'm not sure if related to this change or something else.

arantius:
> Why history.back()?

[...]

PR #2719, opened 2017-11-21
Moved user-script detection to background scripts

Sxderp:
I made these changes in preparation for some other things I was testings some time ago, but I recently found out that moving script detection to the background can help bypass some of the current CSP issues Greasemonkey is having with installing scripts.

Per #2631, script detection fails for CSP hosts. With this change scripts are properly detect.

Some things to note.
1. This makes use of [filterResponseData](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/webRequest/filterResponseData) which is FF 57+ only. So no FF 52 ESR support nor Chrome support (I think Chrome support was dropped/no long a priority per #2560?)
2. Performing the check in the background allows for safely caching a script in the background (in some variable or a database) so when ready for install the extra XHR to retrieve it is unnecessary.


arantius:
This is missing mime type detection, necessary for things like not breaking browsing to the UI view of a script that is stored in e.g. GitHub.

Also, what happens if the server is slow to respond?  What if it's a big script, and it takes much longer to load the whole thing than just the metadata which we strictly need?

Sxderp:
> This is missing mime type detection, necessary for things like not breaking browsing to the UI view of a script that is stored in e.g. GitHub.

I'll have to look at what can be done about extracting the mime type.

> Also, what happens if the server is slow to respond? What if it's a big script, and it takes much longer to load the whole thing than just the metadata which we strictly need?

I'm not sure I understand the concern. With the current method the script is fetched and downloaded, the content script runs, and then we get the install dialog. The whole script has to be loaded anyway.

arantius:
> With the current method the script is fetched and downloaded, the content script runs, and then we get the install dialog. 

Exactly.  There could be a very long delay between clicking the link and seeing anything happen.

> The whole script has to be loaded anyway.

But it doesn't have to be loaded now, while the user waits with no feedback.  Look at how 3.x acted.  It stops as soon as it has the metadata, pops the install dialog, *then* starts the actual download of the script and all resources, while the user can read the dialog and decide what to do.

Sxderp:
From what I understand of the 3.x code flow, this is what happens.
1. If URI matches a user script profile then initiate download of the text content of the script.
2. The entire text content is fetched[1].
3. The dialog box is opened while the dependencies are downloaded in the background.
4. Based on what the user does.
   1. If the user does not install everything is cleaned up.
   2. If the user installs then everything is saved.

At the moment I feel doing significant changes to the install workflow (from the current WebExtension version) is out of the scope of this particular pull request, and should be addressed in followup commits.

[1] There're stubs for what looks like downloading just enough for the metablock but it doesn't seem to be implemented.

arantius:
I used to have a public demo of this but it's broken so launch this: https://gist.github.com/arantius/3b493cb7ae6aad92d6c93881c14f1c4f with GM 3.x installed.  Browse to http://localhost:8000/ and you'll see it take ~10 seconds to load.  Browse to http://localhost:8000/slow.user.js and you'll see that it takes about 2 seconds to pop the install dialog (which then continues for another ~8 to complete the download).

GM 4.x definitely does not do this today (Firefox does some funny buffering of the entire document before it displays any of it, and we don't run until that point anyway).  But that's because 4.x was the simplest hack I could throw together that would work at all.  We should be heading in this direction (UI feedback as soon as possible), not away from it.

Sxderp:
> This is missing mime type detection

Added in recent commit.

> GM 4.x definitely does not do this today 

Added in recent commit. However, still using the full tab for the install message. But, it can be easily switched over when the dialog windows are working. Which, they seem to be working[1] fine in FF 59. Not sure about 57-58.

[1] Title prefix is still semi-broken.

Sxderp:
Oh, I failed to mention that this doesn't affect the current installation flow. That should probably be done in a followup.

Sxderp:
The changes made in this branch partially put us toward the 3.x functionality. If this gets merged I have some additional commits that'll bring the download and install much closer to 3.x. Key points on what I have.

- Pops install as soon as possible (only item on this list that is in this pull request).
- Uses a install dialog / window (works-ish on FF59, as mentioned before, needs testing for previous versions).
- Continues downloading the main script without interrupting the page request.
- As the main script is downloading it also asynchronously starts downloading the dependencies.
- Displays the main script in the original tab, allows for viewing of the source should someone desire.
- If the 'Install' button is pressed the dialog window closes. Resources continue downloading in the background and on completion and database transaction notifies the user (browser.notification) that the script has installed.
- If the 'Cancel' button is pressed the dialog window closes. All the network requests are aborted, save for the main script in the original tab.
- Further, how scripts are updated (from editor save or to be implemented auto-update, the code path should be future proof..ish) was modified too. Should help alleviate #2683.

I could push these changes to this branch, but it's kind of a lot (all having to deal with the same thing though).

Note, there are no progress indicators / bars. I used `fetch` because of their inherent promise-ness and using them makes it very difficult to display some sort of status updates.

Sxderp:
Rebased to deal with #2641. Also includes a small fix to cause the page to close after clicking install.

manifest.json
```diff
@@ -9,7 +9,7 @@
   "applications": {
     "gecko": {
       "id": "{e4a8a97b-f2ed-450b-b12d-ee082ba24781}",
-      "strict_min_version": "52.0"
+      "strict_min_version": "57.0"
```
arantius: Why?

src/bg/on-message.js
```diff
@@ -14,9 +14,7 @@ chrome.runtime.onMessage.addListener(function(message, sender, sendResponse) {
     return;
   }
 
-  if (message.name == 'OpenInstallDialog') {
-    // No-op; content can legitimately ask for the dialog to be opened.
-  } else if (
+  if (
```
arantius: Should be one line now.

src/bg/user-script-detect.js
```diff
@@ -0,0 +1,103 @@
+/*
+Add listeners to detecting user scripts and opening the installation
```
arantius: This form:

    /* Add listeners to detect user scripts and open the installation dialog. */

Is better grammar and fits on one line.

src/bg/user-script-detect.js
```diff
@@ -0,0 +1,103 @@
...
+dialog.
+*/
+
+
```
arantius: Why two blank lines here?

src/bg/user-script-detect.js
```diff
@@ -0,0 +1,103 @@
...
+   'application/x-javascript',
+   'text/ecmascript',
+   'text/javascript',
+]);
+
+
+// Examine headers before determining if script checking is needed
+function checkHeaders(responseHeaders) {
+  for (header of responseHeaders) {
+    if ('Content-Type' === header.name && userScriptTypes.has(header.value)) {
```
arantius: These should be case-insensitive matches, as the server is not guaranteed to provide any particular case for these values.  (Just tested, when I send all-lower values, the header name is exactly "content-type" here.)

src/content/install-dialog.js
```diff
@@ -3,8 +3,8 @@ document.title = details.name + ' - Greasemonkey User Script';
 
 
 function finish() {
-  chrome.windows.getCurrent(null, win => {
-    chrome.windows.remove(win.id);
+  chrome.tabs.getCurrent(curTab => {
+    chrome.tabs.remove(curTab.id);
```
arantius: 1) Why?
2) Desktop vs. Android?

arantius:
Ideally this would act like 3.x, if the load is detected as a script, abort and never show the navigation.  If necessary, always abort (pause?), and when detected _not_ a script, re-start (unpause?).

Right now after patching, I'm seeing both the install dialog pop and the browser navigate to the script.

manifest.json
Sxderp: [filterResponseData](https://github.com/Sxderp/greasemonkey/blob/3c00f700159212296dcbf968bfa043ca0af3af6c/src/bg/user-script-detect.js#L49) is FF 57+
https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/webRequest/filterResponseData

src/content/install-dialog.js
Sxderp: Using tabs works on both devices. Android needs to use tabs since it doesn't have windows. On desktop since the installer is opened using a dialog the window is considered a single tab window. When that tab is removed so is the window.

Sxderp:
> Right now after patching, I'm seeing both the install dialog pop and the browser navigate to the script.

It was so that users could see the source before installing. As mentioned in another issue you talked about the possibility for a malicious server to send different content on the second request that GM currently uses to fetch and save a script. While that concern is valid, it's also somewhat moot, ***at this moment***, due to GM still not having a way to install scripts in a disabled state.

Furthermore, a followup PR that I'd like to submit would resolve the above issue anyway, as it reworks the script download / installation so that a secondary request isn't required, among other things (background / async).

I can look into modifying this PR so that it reflects the abort / navigation redirect that 3.x does. But I don't think, in the longer term, it should remain like that.

Sxderp:
Ah, I don't believe I address #2643 in this PR. Should be an easy fix to add a check at the top of the main function call.

Sxderp:
> I can look into modifying this PR so that it reflects the abort / navigation redirect that 3.x does. But I don't think, in the longer term, it should remain like that.

Attempting to implement the 3.x behavior and I've hit a problem. With a `StreamFilter` the request content is given to us as Firefox receives it. Now, after we have 'enough' content to open an installation dialog I can tell Firefox to stop processing the request.

However, we now have a problem. The user is then presented with a blank page, or a page with a partially downloaded script if `filter.write()` is used before closing. I'm having trouble trying to figure out a 'nice' way to deal with this result.

I can get the tab that the script request belongs to but unfortunately, as far as I can see, there's no `history.back()` equivalent in the background. And I can't just close the tab as it may have a non-zero history. Redirecting using `details.originUrl` is also a poor choice. If the script was opened using middle-click (open in new tab) then `details.originUrl` will be that of the previous tab. After redirect you're stuck with two tabs at the same url.

I also attempted to do a `tabs.executeScript()` with a `history.back()`, but that doesn't seem to be working with a middle-click either.

If this is _absolutely_ necessary, then I'm stumped. Maybe someone else has some ideas?

arantius:
I'd still prefer that clicking a link on a script pops the install dialog and otherwise leaves the browser unchanged, like 3.x.  But this fixes installs from sites with CSP like GitHub and also #2643 both of which are good.  I'm merging it, we can figure out more improvements in the future.

> Note, there are no progress indicators / bars. I used fetch because of their inherent promise-ness and using them makes it very difficult to display some sort of status updates.

I'd also like to see progress bars come back.

arantius:
> (I think Chrome support was dropped/no long a priority per #2560?)

No, the idea was to switch our code to Promise-y `browser` APIs, and use the linked polyfill to get  Chrome compatibility.   (Firefox first, not Firefox only.)  But also supporting Chrome adds a lot of testing/verification burden so who knows if/when it will happen.

Issue #2599, opened 2017-09-29
Cannot use GM.xmlHttpRequest with CORS servers.


ToadKing:
Trying to set the `Origin` and `Referer` headers do not work, so trying to use them with servers protected with CORS fails.

arantius:
A reduced test case, like perhaps a given `.user.js` and a (Python?) snippet to launch a server that should work but fails, would be really helpful.

ToadKing:
Script:

```javascript
// ==UserScript==
// @name        CORS issue
// @namespace   cors.issue
// @include     *
// @grant       GM_xmlhttpRequest
// @grant       GM.xmlHttpRequest
// ==/UserScript==

function loadUrl(url, cb) {
  var options = {
    method: "GET",
    url: url,
    headers: {
      "Origin": "example.com",
    },
    onload: function(res) { cb(res.responseText); },
    onerror: function() { cb(res.responseText); }
  };
  
  if (GM != null) {
    GM.xmlHttpRequest(options);
  } else {
    GM_xmlhttpRequest(options);
  }
}

loadUrl("http://localhost:65432/", function(data) {
  if (data == null) {
    alert("returned an error");
  } else {
    alert("returned \"" + data + "\"");
  }
});
```

Server (NodeJS):
```javascript
const ORIGIN = 'example.com';
const http = require('http');

http.createServer((req, res) => {
    // Set CORS headers
    res.setHeader('Access-Control-Allow-Origin', ORIGIN);
    res.setHeader('Access-Control-Allow-Methods', 'OPTIONS, GET');
    res.setHeader('Access-Control-Allow-Headers', '*');

    if (req.method === 'OPTIONS') {
        res.writeHead(200);
        res.end();
        return;
    }

    if (req.headers.origin != ORIGIN)
    {
        res.writeHead(403);
        res.write("not allowed");
        res.end();
        return;
    }

    res.writeHead(200);
    res.write("ok");
    res.end();
}).listen(65432);
```

The alert the userscript returns should say "ok" but instead returns "not allowed".

DoctorMcKay:
+1, cannot set `Referer` header.

NoneGiven:
@arantius This is annoyingly enforced by Firefox per [the XHR spec](https://fetch.spec.whatwg.org/#forbidden-header-name) which defines Referer (and some others) as a forbidden header. I was able to work around it in my own WebExt like this:

```js

// background-script.js

function rewriteHeaders(e) {
[...]

PR #2723, opened 2017-11-22
Allow user to set forbidden headers in XHR

NoneGiven:
Even in WebExtensions, Firefox enforces [per spec](https://fetch.spec.whatwg.org/#forbidden-header-name) that Referer, Origin, and other headers may not be set by the developer on the XHR object. However, if the outgoing XHR is intercepted in the `onBeforeSendHeaders` event, it is possible to set these headers as desired.

~~This is a WIP~~ This change currently looks for Referer and Origin and shows how GM can set the header values when the user supplies them the normal way, through the `headers` property of the object passed to `GM.xmlHttpRequest`. It's necessary for GM to add dummy headers to preserve the user's header values until `onBeforeSendHeaders`, at which point we can set the real header and delete the dummy.

Fixes https://github.com/greasemonkey/greasemonkey/issues/2599

### DIFF
--- a/src/bg/on-user-script-xhr.js
+++ b/src/bg/on-user-script-xhr.js
@@ -99,12 +99,12 @@ function open(xhr, d, port) {
   d.overrideMimeType && xhr.overrideMimeType(d.overrideMimeType);
   d.responseType && (xhr.responseType = d.responseType);
   d.timeout && (xhr.timeout = d.timeout);
-
+  
   if (d.headers) {
     for (var prop in d.headers) {
       if (Object.prototype.hasOwnProperty.call(d.headers, prop)) {
         var propLower = prop.toLowerCase();
-        if (propLower in headersToReplace) {
+        if (headersToReplace.includes(propLower)) {
           xhr.setRequestHeader(dummyHeaderPrefix + propLower, d.headers[prop]);
         }
         else {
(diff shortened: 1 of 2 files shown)


NoneGiven:
One thing that needs investigation is whether the `<all_urls>` in the listener registration can hopefully be replaced with `*://*/*.user.js`. We don't actually want GM running `rewriteHeaders` for every single request the browser makes.

xor10:
Use `originUrl` to check if requests are from extension. :)
Should be allowed to modify the `Host` and `User-Agent` headers too.

```javascript
const extensionUrl = chrome.extension.getURL('');
function rewriteHeaders(detail) {
  // rewrite only requests from background script (GM.xmlHttpRequest)
  if (detail.originUrl && detail.originUrl.startsWith(extensionUrl)) {
    const newHeaders = [];
    ...

    return { requestHeaders: newHeaders };
  }

  return { requestHeaders: detail.requestHeaders };
}

```

NoneGiven:
Thanks @xor10 ! I added the originUrl check and also rewrote the code to actually work (had some confusion over which objects were plain objects and which were arrays). I tested this with the modified GM code loaded into FF 58 and it works great.

Setting `User-Agent` is already possible. For the other two, maybe @arantius can decide on the final list of forbidden headers to support.

src/bg/on-user-script-xhr.js
```diff
@@ -119,4 +127,37 @@ function open(xhr, d, port) {
...
+
+var extensionUrl = browser.extension.getURL("");
+
+function rewriteHeaders(e) {
+  if (e.originUrl && e.originUrl.startsWith(extensionUrl)) {
+    for (var name of headersToReplace) {
+      var prefixedHeader = getHeader(e.requestHeaders, dummyHeaderPrefix + name);
+      if (prefixedHeader) {
+        if (!getHeader(e.requestHeaders, name)) {
+          // only try to add real header if request doesn't already have it
```
Sxderp: I'm not sure I understand the purpose of this conditional, could you provide an example of when `getHeader` may return true on line 147?

Further, under what condition would you _not_ want to replace a header with the prefixed version?

NoneGiven: It would return true for headers like `Accept-Encoding` and `Connection` that are always automatically set, if we were allowing the user to replace those.

>under what condition would you not want to replace a header with the prefixed version?

I guess you wouldn't.

NoneGiven:
@arantius Anything I can do to move this PR along? Unlike that other one, this one works and has a legitimate goal (giving `GM.xmlHttpRequest` the same utility as the old `GM_xmlhttpRequest`).

arantius:
Sorry!  Merging now.

PR #2732, opened 2017-11-28
Include query string in MatchPattern URL test

### LINKED ISSUE #2715: @Match problem (not ports)
Just noticeed that after migration to new FF and GM4 my match:
`// @match		*://*.gelbooru.com/index.php?page=post&s=view&id=*`
no longer fires on url (example)
https://gelbooru.com/index.php?page=post&s=view&id=2865115

while simple `// @match		*://*.gelbooru.com/*`
fires on that url

was working perfectly in GM3, but in GM4 it has this weird bug... 

whole script if you wonder: https://github.com/Owyn/HandyImage/raw/master/HandyImage.user.js

NoneGiven:
Super simple fix prescribed by @arantius, fixes #2715.

Tested with this branch loaded as a temporary webext in FF 57 and it does indeed fix the issue (obviously).

### DIFF
--- a/test/bg/user-script-registry.test.js
+++ b/test/bg/user-script-registry.test.js
@@ -31,27 +31,4 @@ describe('bg/user-script-registry', () => {
       });
     });
   });
-
-  it('match pattern respects query string', async () => {
-    let userScriptWithQuery = new EditableUserScript(
-        {'name': 'highlight',
-         'matches': ['https://example.com/test?query=*'],
-         'content': 'void(0)'});
-    await UserScriptRegistry._saveUserScript(userScriptWithQuery);
-    let userScriptNoQuery = new EditableUserScript(
-        {'name': 'lowlight',
-         'matches': ['https://example.com/test'],
-         'content': 'void(0)'});
-    await UserScriptRegistry._saveUserScript(userScriptNoQuery);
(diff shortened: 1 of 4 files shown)


arantius:
Please include a unit test.

NoneGiven:
Added unit test.

```
Firefox 57.0.0 (Windows 7 0.0.0): Executed 28 of 28 SUCCESS (0.617 secs / 0.585 secs)
```

test/bg/user-script-registry.test.js
```diff
@@ -31,4 +31,27 @@ describe('bg/user-script-registry', () => {
       });
     });
   });
+
+  it('match pattern respects query string', async () => {
```
arantius: Instead of `it('match pattern ...`, this should be a test of MatchPattern.doMatch() directly.  That's the code you're changing.  And it would also need less setup code, tested that way.

NoneGiven: Should I add a `test/third-party` directory and `MatchPattern.test.js` then?

arantius: Yes

NoneGiven:
Added `doMatch` unit test.

arantius:
Great!  Thanks for all the updates!

PR #2734, opened 2017-11-28
Sync both inactive and active scripts on uninstall

Sxderp:
Currently when you uninstall a script and go back to the top of the Monkey Menu, nothing changes. This will sync both inactive and active scripts.


Sxderp:
I don't believe this is necessary due to recent changes to the monkey menu.

PR #2743, opened 2017-12-03
Fix logical flaw from 7d1f1b98

Sxderp:
The initial change fixed the problem but only as a side effect that
typeof returned 'object' for everything. Therefore any object would
fail the comparison and get caught.


arantius:
Tests to cover all cases would be an extremely welcome addition.  (If we did it the first time, we never would have had this bug!)

Sxderp:
Sure, I'll write some. Even though they're just some simple tests I'm not sure if I'll be able to eek them out until after Tuesday. Big stuff going on Monday + Tuesday.

arantius:
I found time to write the tests.  Still thanks for your contribution!

Issue #2714, opened 2017-11-20
Reconsider implementing GM_registerMenuCommand in GM 4.x


darkred:
As I saw (see comment) `GM_registerMenuCommand` is replaced with HTML5 context menu. and there's no plan to add it in GM 4.x.

Nevertheless, I've started this separate issue. in hope that you'll might reconsider.  

---

I understand that adding some context menu options in the native context menu, 
is useful for those that do want to add more entries in that.
Such as this new test script by arantius [search-with-google](https://arantius.com/misc/greasemonkey/search-with-google.user.js).
screenshots:
![2017-11-20_171908](https://user-images.githubusercontent.com/723651/33025536-f4d5e1bc-ce16-11e7-85c1-4ca81cde1c55.jpg) ![2017-11-20_172014](https://user-images.githubusercontent.com/723651/33025608-228be46c-ce17-11e7-98c6-6270125c6704.jpg)

<br/>

But, for userscripts that use `GM_registerMenuCommand` only for settings, i.e. you only configure then only once in a while.
and apply to all pages, 
such as the popular:
[Mouseover Popup Image Viewer](https://greasyfork.org/en/scripts/404-mouseover-popup-image-viewer), ([settings screenshot](https://greasyfork.org/system/screenshots/screenshots/000/000/367/original/settings.png?1425674740))
[YouTube Link Title](https://greasyfork.org/en/scripts/413-youtube-link-title) ([settings screenshot](https://user-images.githubusercontent.com/723651/33026925-57df9232-ce1a-11e7-9243-c72238b3f9f4.jpg)) and 
[Linkify Plus Plus](https://greasyfork.org/en/scripts/4255-linkify-plus-plus) ([settings screenshot ](https://user-images.githubusercontent.com/723651/33026930-5bd775bc-ce1a-11e7-9561-21dc9efd026d.jpg))

Imagine having installed these scripts that create a total of 3 entries on top of the context menu
that are not needed most of the time:   
it would make use the context menu on the page cluttered and unpractical.

And then the script author, in order to avoid cluttering the context menu just for the sake for offering settings,
he would either register a keyboard shortcut on every page (with no UI?)
or create a new dedicated button element on every page.

<br/>

So, I'd like to ask you please to reconsider implementing `GM_registerMenuCommand` inside the toolbar button popup, similar to how it was in GM 3.x, as well as in the other most used scripts managers available, Tampermonkey and Violentmonkey. 
[...]

PR #2770, opened 2017-12-14
Add GM.registerMenuCommand()

esperecyan:
Resolves https://github.com/greasemonkey/greasemonkey/issues/3078 , resolves https://github.com/greasemonkey/greasemonkey/issues/3099 , and resolves one part of https://github.com/greasemonkey/greasemonkey/issues/2714 .

Demo
=====
First, we click Greasemonkey toolbar button to open menu. Then, we select a string “User script commands ...” to display command caption strings.

Demo Script
=========
```javascript
// ==UserScript==
// @name     GM.registerMenuCommand() Demo
// @version  1
// @match    https://github.com/greasemonkey/greasemonkey/*
// @grant    GM.registerMenuCommand
// @grant    GM.openInTab
// ==/UserScript==

GM.registerMenuCommand('Alert', () => alert('[A]'));
GM.registerMenuCommand('Open GitHub', () => GM.openInTab('https://github.com/greasemonkey/greasemonkey/', false), 'G');
GM.registerMenuCommand('Open wiki', () => GM.openInTab('https://wiki.greasespot.net/', false), 'w');
```


Sxderp:
Gave it a brief look. Like the thoroughness of the test. Off the top, I don't think I saw anything about duplicate access keys (may have missed it).

Secondly, `doc/Messages.md` should be modified with 'ListMenuCommands' and 'MenuCommandClick'.

esperecyan:
@Sxderp Thank you for pointing those out. I overlooked those. I revised to
- make commands having duplicate access key focus sequentially
- modify doc/Messages.md
- simplify the tests
- make back buttons CSS same

esperecyan:
I resolved conflicts with newer commits.

esperecyan:
I resolved conflicts with aaadd6f15aabf1da87b9690cf2f1aa9d3539c882 and 8b5b6507d7961ad18dc184cfef42572b10acc1ba and fixed code style.

Sxderp:
Skimmed the code, looks okay, I haven't tested the updates.

---

Although, unrelated to the PR, this makes me think that `monkey-menu.js` needs to be refactored a bit. Split up into smaller functions (like the giant case statement), or even split into additional files. Furthermore, as additional sub menus are added it might be better to make the selection objects a single JS object with properties for each sub menu, rather than individual variables (again, for some other PR / commit).

esperecyan:
Thank you for checking.

I also think that code reorganization and splitting into additional files is necessary.
About keyboard navigation, how about using `document.activeElement` and `Element#closest()` (and then use querySelectorAll) to find next menu item?
https://developer.mozilla.org/docs/Web/API/Document/activeElement
https://developer.mozilla.org/docs/Web/API/Element/closest

Sxderp:
I feel like that might be a lot of overhead for something that occurs fairly frequently. Sure it's small in comparison to everything but I don't see the gains in maintainability being significant. Unless I'm missing something you're talking about fetching the list of `a` tags based on your current location in the DOM?

esperecyan:
I suppose that the way has not overhead causing problems but I have no objection to holding list.

> Unless I'm missing something you're talking about fetching the list of a tags based on your current location in the DOM?

I thought so, but in retrospect, using `Element#closest()` was improperly because I did not consider patterns that there is `document.activeElement` out of displayed section.

esperecyan:
I resolved conflicts with https://github.com/greasemonkey/greasemonkey/commit/cfd74d716e8351688be9114f663471196e4122fb .

Sxderp:
If you add the following to the css

```css
.subview-item.has-shortcut::after {
  content: attr(aria-keyshortcuts);
}
```

Then add 'has-shortcut' to the `<a>` tag [class](https://github.com/greasemonkey/greasemonkey/pull/2770/files#diff-328223e7364c9a5bee7c926bf6cb8bf6R139), you should be able to get the automatic shortcut formatting as shown in the Monkey Menu PR [1]. Which should remove the need for the `rv-if` as well.


[1] https://github.com/greasemonkey/greasemonkey/pull/2805#issuecomment-357571769

esperecyan:
@Sxderp Thank you. I made the shortcut keys displayed like the FF Menu Bar because they can be used only inside menu and an item on a native FF menu panel seems to display a shortcut key that can be used also outside the menu.

esperecyan:
I resolved conflicts with https://github.com/greasemonkey/greasemonkey/commit/9ce5dabc74496fa6a631d991302ad4fd80730f1d .

esperecyan:
I resolved conflicts with https://github.com/greasemonkey/greasemonkey/commit/3f4e1c53731cdc8196b760f9cb59fb52921242f6 , https://github.com/greasemonkey/greasemonkey/commit/5ff1486c4a4a0e61f76964a01d1d21bab65dcad6 , and https://github.com/greasemonkey/greasemonkey/commit/1ba6d09df7057bf76e8e921183db759b00c7eae6 .

_locales/en/messages.json
```diff
@@ -32,6 +34,8 @@
   "This is a Greasemonkey user script.  Click install to start using it.": { "message": "This is a Greasemonkey user script.  Click install to start using it." },
   "Undo uninstall": { "message": "Undo uninstall" },
   "Uninstall": { "message": "Uninstall" },
+  "User script commands": { "message": "User script commands" },
+  "User script commands ...": { "message": "User script commands ..." },
```
Sxderp: Since these are the exact same message, I don't think it's required to have both. Perhaps just adding the "..." in the code, while keeping the same translation message?

Maybe this is just an English thing and other languages would have different phrases, I don't know.

esperecyan: `User script commands` is a heading of the sub menu and `User script commands ...` is a label of the top menu item.

I also don’t know, but I added the latter one in accordance with `New user script ...`.

Eselce: The text is the same, the usage may be not. Do you really want to split `Undo uninstall` into `Undo`, `SPACE` and `toLowercase`(`Uninstall`) in order to avoid same-text-translations? (NOT)

This may be exagerated, but the point is, these strings may be different in a translation.
Another example: Some languages conjugate differently. Some are associative.

Last example: Here I'd see even english "translations", that may differ:
Heading of sub menu may read `List of user commands`, while the main menu item could be `User script commands...`.

BTW: I'd prefer no `SPACE` before `...`, as seen in merely every other `Firefox` menu...

arantius:
Sorry for delays.  Today I checked this out and merged master in; I installed a single script that calls this API:

```javascript
// ==UserScript==
// @name     GM.registerMenuCommand test
// @version  1
// @grant    GM.registerMenuCommand
// ==/UserScript==

GM.registerMenuCommand(
    'GM_rmc works', function(){ console.log('GM_rmc really works.'); });
  ```

And then after loading a couple pages (clicking links in one tab) I got:

untitled

The same command registered four times.  I clicked on each one and each one caused the menu to close, but only the last one caused the message to be logged.  (I'd also prefer the script's icon, over the same generic "command" (?) icon for every item.  No icons would be better than that, actually.)


This is basically exactly why the delay.  I've long hated GM_rmc because it's _so hard_ to get it right.  In this case, I _think_ the back/forward cache is keeping the last three history items alive, so the port isn't disconnected?  Throwing some logs in seems to confirm this.  On some other pages it is, but not always.  (Test failed at [this](https://gist.github.com/arantius/1157543#file-infinite-php).)

Also, smaller but still an issue, I now see the tests log every time:

    WARN: 'Missing i18n str:', 'GM.registerMenuCommand: "commandFunc" must be a function'
    WARN: 'Missing i18n str:', 'GM.registerMenuCommand: "accessKey" must be a single character'
    WARN: 'Missing i18n str:', 'GM.registerMenuCommand: "accessKey" must be a single character'
    WARN: 'UserScriptMenuCommand port un-handled message name:', 'fakeName'

File: _locales/en/messages.json (hunk 1 above)
Comment (arantius) [reply _locales/en/messages.json]: > BTW: I'd prefer no SPACE before ..., as seen in merely every other Firefox menu...

+1

**Comment by Sxderp** [comment]
>  I think the back/forward cache is keeping the last three history items alive, so the port isn't disconnected?

If so, then related to #2705. The _best_ solution that I can see is to let Firefox handle script execution through the `contentScripts` (59) or `userScripts` (??) API. Perhaps this PR should wait until then? The page caching already bites us, and this makes it bite us harder.

**Comment by Eselce** [comment]
Can't wait to see GM.rmc succeed... ;-)

File: _locales/en/messages.json (hunk 1 above)
Comment (esperecyan) [reply _locales/en/messages.json]: What should we do about `New user script ...`?

Comment (Sxderp) [reply _locales/en/messages.json]: @esperecyan Submit a pull.

**Comment by Sxderp** [comment]
> I'd also prefer the script's icon, over the same generic "command" (?) icon for every item. No icons would be better than that, actually.

I think the script icon would be better. At least then you can, assuming the script has an icon, at a glance determine what script a command belongs to. Of course that doesn't work for scripts that don't use an icon, but you lose nothing since you're in the same boat as if there were no icons at all.

**Comment by esperecyan** [comment]
```
WARN: 'Missing i18n str:', 'GM.registerMenuCommand: "commandFunc" must be a function'
WARN: 'Missing i18n str:', 'GM.registerMenuCommand: "accessKey" must be a single character'
WARN: 'Missing i18n str:', 'GM.registerMenuCommand: "accessKey" must be a single character'
WARN: 'UserScriptMenuCommand port un-handled message name:', 'fakeName'
```

I suppressed the messages in test, but I see `ReferenceError: _ is not defined` when codes such as `GM.registerMenuCommand()` or `GM.notification()` are run in pages.

And then, I made menu show script icons.

**Comment by Eselce** [comment]
???

`function _(...)` is defined in `src/i18n.js`, including the first three `WARN`ings on the console...
This function returns the untranslated string, if it cannot find a translation (with `WARN`)!

**Comment by Eselce** [comment]
The last (forth) message results from this (only 'register' is allowed)
```
      fakePort.onMessage.trigger({
        name: 'fakeName',
      }, fakePort);
```

**Comment by esperecyan** [comment]
@Eselce Oh, <q>I suppressed the messages in test</q> mean that I fixed <q>Also, smaller but still an issue, I now see the tests log every time:</q>

`ReferenceError: _ is not defined` is seen in a script such as:

```js
// ==UserScript==
// @name     i18n test
// @version  1
// @grant    GM.notification
// ==/UserScript==

try {
	GM.notification();
} catch (exception) {
	alert(exception);
}
```

Eselce:
I see. Shouldn't you give at least a notification text?

esperecyan:
Yes. I do not know the best way how to modify `_` function call in a content script context.

Eselce:
That's probably a "thing" for most of the internal messages (I remember @arantius leaving them alone)...

Eselce:
> That's probably a "thing" for most of the internal messages (I remember @arantius leaving them alone)...

I did a check: There are a handful occurences of such error message, and it's possible to fix it (thinking about a PR)...

arantius:
> ReferenceError: _ is not defined is seen in a script such as:

Whoops, oversight on my part.  The current implementation does some dirty `.toString()` tricks -- so the source can be in a file with normal editor syntax highlighting.  But it means it leaks from what appears to be extension context into user script context.  So we need to either define `_` in the user script API handling context, or do translations another way for error messages in those API handlers.

(Ideally, the strings would probably be looked up once, when the user script source is generated.)

Eselce:
@arantius "Either" is done in #2840

esperecyan:
I resolved conflicts with https://github.com/greasemonkey/greasemonkey/commit/954107b76cd7ae1f83ace8c160ade2656c97dbb4 , https://github.com/greasemonkey/greasemonkey/commit/65f2912783f0ec15f4bd9a5cffee752fd7c21cec , and https://github.com/greasemonkey/greasemonkey/commit/0fb3c9776b2c7d7bfb0b1045ffdcea26ea4c042c .

Sxderp:
@esperecyan was https://github.com/greasemonkey/greasemonkey/pull/2770#issuecomment-363110990 resolved? Or are we waiting for contentScript API?

esperecyan:
@Sxderp I made `Port#disconnect()` be called when a window fires unload event to resolve it now.

poke:
I would love to see this feature make it into GM! Is there anything we could help with to get this moving forward again?

esperecyan:
I adjusted to the new menu.

arantius:
I've merged this.  I'd appreciate if anyone concerned could test build https://addons.mozilla.org/firefox/downloads/file/3711152/greasemonkey-2021.1.18.15493nightly-an+fx.xpi before we cut a release.

Cerberus-tm:
In the latest nightly, the Demo Script above seems to work perfectly! Thanks @esperecyan for the great work.

poke:
I’ve wanted to test this too for a few of my scripts where I am currently drawing buttons into the document. The nightly link above however no longer works. Is there an updated nightly release I can test without building the plugin myself?

Cerberus-tm:
@poke Arantius posted another link, where he hosts the nightly himself:
https://github.com/greasemonkey/greasemonkey/pull/3090#issuecomment-762474380

poke:
@Cerberus-tm Thanks a lot!

So, I’ve tested this now and it appears to work perfectly fine. Commands are correctly registered and selecting them also triggers the actions correctly. – Ideally, I would like to promote the user script commands into the top-level menu so that you only need two clicks (instead of three) to run a command, maybe by moving the “Other user scripts” into a submenu instead. But I can totally live with this for now!

One issue that I have found with the current Nightly is that the menu appears to be cut off due to the new item in it:

Menu is cut off at the bottom

If you hover over the cut-off menu item “Get user scripts” at the bottom, then the menu appears to scroll down, cutting off the top instead:

Menu is cut off at the top

Cerberus-tm:
Ah, I see it too.

Purfview:
> 
> 
> I've merged this. I'd appreciate if anyone concerned could test build https://addons.mozilla.org/firefox/downloads/file/3711152/greasemonkey-2021.1.18.15493nightly-an+fx.xpi before we cut a release.

I tested "greasemonkey-2021.01.18.nightly.7409ca7c.xpi" on Firefox v85, I see entry at "User script commands..." but it doesn't do anything when pressed, nothing in Browser Console.

`GM.registerMenuCommand('IMDb Scout Mod Settings', function() {GM_config.open()});`

https://github.com/Purfview/IMDb-Scout-Mod/issues/10

Cerberus-tm:
@Purfview That version I cannot download: the page gives an error. But i am using Arantius's self-hosted nightly from here:
https://github.com/greasemonkey/greasemonkey/pull/3090#issuecomment-762474380
It works well for me. I've tried using this code in another script as a test, and the message is printed in the console as expected for me. Are you sure it isn't due to some issue in your GM_config class/object?
`GM.registerMenuCommand('IMDb Scout Mod Settings', function() {
	console.log('Hello')
});`

Purfview:
@Cerberus-tm 

That's the same version I tried ("greasemonkey-2021.01.18.nightly.7409ca7c.xpi").

`GM_config.open()` is from []https://github.com/sizzlemctwizzle/GM_config, tested with other monkeys, no issues there.

poke:
@Purfview The problem with that script appears to be unrelated to this feature about script commands though. I have tested quite a few different scripts that do non-trivial stuff when a menu command is selected and they all work nicely. So I would consider this feature/PR successful.

Note that GM 4 behaves very differently compared to Tampermonkey or Violetmonkey. So the reason why that script doesn't work with GM could be very different. The fact that you owned that issue there a few weeks ago also makes it very clear that the issue is not related to this feature (which was only merged a week ago and hasn't been released yet).

Issue #2672, opened 2017-11-15
GM.openInTab opens the tab in the wrong place


AndersOhrt:
I've upgraded to Firefox 57, and migrated my script from GM_openInTab to GM.openInTab. I have browser.tabs.insertRelatedAfterCurrent = false. When tabs are opened, they are placed incorrectly, as if insertRelatedAfterCurrent was true.

xor10:
Webextensions don't have access to this setting. :(
The solution would be to create a settings page to allow the user to set this.

AndersOhrt:
Adding another parameter to the function would also work in my case, as it is now my script is worthless.

PR #2772, opened 2017-12-16
Make `GM.openInTab` open a tab in a correct place in Firefox 57+.

esperecyan:
Fixes https://github.com/greasemonkey/greasemonkey/issues/2672 .

This uses `openerTabId` property of `tabs.create()` that implemented in Firefox 57.
https://developer.mozilla.org/Add-ons/WebExtensions/API/tabs/create

### Script used by test

```javascript
// ==UserScript==
// @name     GM.openInTab() Test
// @version  1
// @match    https://github.com/greasemonkey/greasemonkey/issues/2672
// @grant    GM.openInTab
// ==/UserScript==

const parent = document.getElementsByClassName('gh-header-show')[0];
parent.insertAdjacentHTML('beforeend', `
  <button name="open-in-background-tab">Open the wiki in background tab</button>
  <button name="open-in-active-tab">Open the wiki in active tab</button>
`);

parent.addEventListener('click', function (event) {
  switch (event.target.name) {
    case 'open-in-background-tab':
      GM.openInTab('https://wiki.greasespot.net/GM.openInTab');
      break;
    case 'open-in-active-tab':
      GM.openInTab('https://wiki.greasespot.net/GM.openInTab', false);
      break;
   }
});
```

### DIFF
--- a/src/bg/on-user-script-open-in-tab.js
+++ b/src/bg/on-user-script-open-in-tab.js
@@ -5,11 +5,18 @@ This file is responsible for providing the GM.openInTab API method.
 function onApiOpenInTab(message, sender, sendResponse) {
   checkApiCallAllowed('GM.openInTab', message.uuid);
   const senderTab = sender.tab;
-  chrome.tabs.create({
+  const tab = {
     url: message.url,
     active: message.active,
-    windowId: senderTab.windowId,
-    index: senderTab.index + 1, // next to senderTab
-    openerTabId: senderTab.id,
+  };
+  chrome.runtime.getBrowserInfo(browserInfo => {
+    if (browserInfo.name === 'Firefox' && browserInfo.version.split('.')[0] < 57) {
+      tab.windowId = senderTab.windowId;
+      tab.index = senderTab.index + 1; // next to senderTab
+    } else {
+      tab.openerTabId = senderTab.id;
+    }
+
+    chrome.tabs.create(tab);
   });
 };


esperecyan:
I resolved conflicts with 8b5b6507d7961ad18dc184cfef42572b10acc1ba and fix code style.

esperecyan:
I changed the PR into completing https://github.com/greasemonkey/greasemonkey/commit/256bf162037639e101bd432f3d7adff5d00a79dd and made this use assignment operators instead of Object.assign().

esperecyan:
@arantius  I am confused about the supported Firefox versions. I was thinking that Greasemonkey 4.x supports Firefox 52 ESR. I think that it can be installed on 52 but the operation is not ensured now. Could you confirm my understanding is correct?

Sxderp:
Is indexeddb still broken in the latests ESR?

esperecyan:
> Is indexeddb still broken in the latests ESR?

As for GM.setValue/GM.getValue, they seem to work correctly at least on my Firefox 52.5.3 on Windows 7 with GM 4.1 and 4.2. 

GM.openInTab is broken at GM 4.2 on Firefox 52.5.3 because 52.x does not support `tabs.Tab.openerTabId`.

Sxderp:
I'd vote for not supporting 52. If someone is on that then they have access to 3.x, which is still more feature-full than 4.x. There's zero reason to switch. 52 tends to create oddities and incompatibilities, as well as not having (by definition) 57+ APIs. I don't see the benefit.

makyen:
@Sxderp I'm assuming that, yes, IndexedDB is still an issue in the . I get the following error messages using GM3.17 in ESR 52.5.3 whenever GM storage is accessed:

    IndexedDB UnknownErr:          ActorsParent.cpp:599  (unknown)
    Error opening script store DB! 0bc1244f-73d6-47ba-821d-c0eac2dc99b7 error { target: IDBOpenDBRequest, isTrusted: true, currentTarget: IDBOpenDBRequest, eventPhase: 2, bubbles: true, cancelable: true, defaultPrevented: false, composed: false, timeStamp: 1516406278086000, cancelBubble: false, originalTarget: IDBOpenDBRequest }                    value-store.js:19:7
    UnknownError                   value-store.js:17:17

I'm assuming the issue only affects the WebExtension side, as values can be stored and retrieved using `GM_setValue()`/`GM_getValue()`.

I haven't tested if the errors result in not being able to use `GM.setValue()`/`GM.getValue()` in GM4.

arantius:
> @arantius I am confused about the supported Firefox versions. I was thinking that Greasemonkey 4.x supports Firefox 52 ESR.

The latest release: no, because `openerTabId` leaked in, I got confused about which changes were merged.  I didn't realize something added that before this.  I think there was also a 57+ only change made recently too, though.

esperecyan:
@arantius Thank you for response. A part of Greasemonkey 4.2 features is not supported on Firefox 52 ESR and whether Greasemonkey 4.3 will support Firefox 52 ESR or not is not decided yet, I gather.

If so, I pretty much agree with @Sxderp. If I must cite a concern, when Greasemonkey 3.x needs to be updated, it is troublesome because Addons.mozilla.org see submission date, not version number.

arantius:
> when Greasemonkey 3.x needs to be updated

This is not something that I ever expect will happen.

esperecyan:
I said that, but I guess that it is not necessary until the EOL of 52 even if 4.x not supports 56 or earlier.

PR #2774, opened 2017-12-19
Add GM.xhr option to use page cookies

### LINKED ISSUE #2555: Make GM_xmlhttpRequest use the cookies from the current Firefox container
Would it be possible to make GM_xmlhttpRequest use the cookies from the current [Firefox container](https://testpilot.firefox.com/experiments/containers/)?

When using the [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch) I can easily accomplish this by using `credentials: "include"`, but GM_xmlhttpRequest appears to always use the cookies from the regular page instead of using the cookies from the current container, leading to unexpected results.

NoneGiven:
Allows `GM.xmlHttpRequest` to send the cookies belonging to the page (that is, belonging to the URL of the tab which is running the content script calling `GM.xmlHttpRequest`). This builds off the request header changes in #2723, so only the last commit of this PR has real changes for this issue.

Since these changes would also give the user the ability to manually specify cookies in the `Cookie` header, I included an explicit option `usePageCookies`. So (1) if the user doesn't set anything, no cookies are sent, (2) if the user enables `usePageCookies`, the page cookies are sent, and (3) if the user manually sets the `Cookie` header, those cookies are sent and the `usePageCookies` setting is ignored. I don't have strong feelings about the precedence here, or even having the option at all, so feel free to suggest a different strategy.

Fixes #2555.


NoneGiven:
The test failure is `ReferenceError: browser is not defined` on `const extensionUrl = browser.extension.getURL('');` sooo...?

Sxderp:
> The test failure is because it wants me to replace all the browser with chrome.

The testing suite was built for Chrome. As a side effect of the move WebExtensions _most_ API calls and whatnot work with Firefox. Unfortunately Chrome does not have the `browser` namespace. I think this is why it fails, the tests are not expecting that namespace.

arantius:
> Allows GM.xmlHttpRequest to send the cookies belonging to the page (that is, belonging to the URL of the tab which is running the content script calling GM.xmlHttpRequest). 

Regardless of which domain is being fetched?

arantius:
> The test failure is because it wants me to replace all the browser with chrome...

https://travis-ci.org/greasemonkey/greasemonkey/builds/318844181

Passes at head, so it's something you've done?   Also, consistency is important when submitting to an existing project; we use `chrome` everywhere.  (For now..)

Sxderp:
> we use chrome everywhere. (For now..)

This particular item doesn't really affect functionality but it would be nice to see the switch. `then` chaining with Promise returns are very convenient and I've tended to utilize them in my personal branches.

arantius:
>> we use chrome everywhere. (For now..)
> This particular item doesn't really affect functionality but it would be nice to see the switch.

Agreed, thus #2560.

Eselce:
Considering #2614 ? Probably yes...

arantius:
I still don't see any response to [my comment above](https://github.com/greasemonkey/greasemonkey/pull/2774#issuecomment-352888500).  To which domain(s) does this send the current page's cookies?

Sxderp:
> I still don't see any response to my comment above. To which domain(s) does this send the current page's cookies?

Firstly, this PR is a bit out of date. It built off of #2723 which has changed. The #2723 one should probably be reviewed first, then after this has been rebased for the changes, this should be reviewed.

Secondly, in the current implementation I think this PR should be a no-go. As far as I can tell the cookies, assuming `includePageCookies` is true, are sent to whatever domain is being `xhr` requested. While cross-domain cookie sharing isn't strictly prohibited (see [fetch](https://fetch.spec.whatwg.org/) and [CORS](https://www.w3.org/TR/cors/)) this particular PR doesn't respect CORS.

With all of that said, I believe this is the intended behavior of the PR. However, as I said I think, as it stands, this is a no-go, and should probably have another permission attached. Though, that may be confusing to users / UI perspective.

Although, I can't think of the use case for wanting to send cookies / credentials / etc to a domain that hasn't been explicitly allowed by the host domain.

NoneGiven:
@arantius @Sxderp Sorry, somehow I never saw that comment despite seeing all the others in this thread.

No, that's an error in my code. We should not be sending cookies for the current tab's URL, but for the domain/URL being fetched. I believe the correct thing is to pass the destination URL to `chrome.cookie.getAll` (and remove the unnecessary tab-related promise), but I'll test and confirm.

NoneGiven:
I'm having one of those days. The issue I opened this PR for is not even about restoring GM_xmlhttpRequest behavior like I thought, it's about the Firefox identities/personas/containers feature. Given how badly I screwed this up, I'm going to close this, let #2723 be reviewed and hopefully merged, and then redo these changes properly then.

NoneGiven:
Note to self for when/if this is revived: to most closely follow the behavior of `GM_xmlhttpRequest` (and have a measure of added security I guess?), check the sender tab URL against the destination domain of the XHR, and only include cookies if they are the same origin.

PR #2778, opened 2017-12-22
Remove CodeMirror flexbox style

badbrainz:
The editor has a layout issue that is affecting typing speed. Removing `display: flex` from the `.CodeMirror` selector should fix this.

### DIFF
--- a/src/content/edit-user-script.css
+++ b/src/content/edit-user-script.css
@@ -36,14 +36,17 @@ html, body {
 }
 
 
-body, #editor, .CodeMirror {
+body, #editor {
   display: flex;
   flex-direction: column;
   flex: 1 1 auto;
   flex-basis: fill;
 
 }
-#editor .CodeMirror {
+#editor {
   height: 0;
   background: white;
+}
+.CodeMirror {
+  height: 100% !important;
 }


arantius:
Thanks for the pull request!  I'll need to test this one before I merge it though.  (And I'm traveling now so ...)

Sxderp:
Tested it. Visually they're the same. However, I don't see any noticeable difference in typing speed. However, my test was writing prose, not code. So either my ~80 words a minute isn't fast enough to trigger this difference or we're not talking about 'just typing'.

badbrainz:
I'm seeing this on Linux. Performance seems to drop when line count exceeds ~50. I forgot to mention that cursor movement is also affected. One thing I noticed is that these issues don't happen when there's a horizontal scrollbar.

I don't know the exact cause but removing `display: flex` does make this go away.

Sxderp:
Yeah, checked again. On Windows and a Linux distro. Still no difference. But since they're visually identical and we have at least one case of improvement ( @badbrainz ), I don't see why it _shouldn't_ be merged.

badbrainz:
So you've tested vertically (holding down <kbd>Enter</kbd> or <kbd>Ctrl+X</kbd>) and horizontally (random char or <kbd>Backspace</kbd>) and still nothing?

arantius:
FWIW I see, at master, no issue scrolling horizontally nor vertically.  No issue typing in that script which is big enough to cause both.  So this has no positive effect (for me), but none negative either, so OK.

arantius:

Issue #2777, opened 2017-12-21
onApi* message handlers have no privilege checking


Sxderp:
If you just copy the code that is normally injected when asking for a `@grant` into the userscript, you can perform the action (send a message to the background) without a problem. Fixed by adding a permission check (against script uuid) for `onApi*` messages in `on-message.js`, or for the case of notification (which uses a port) in the port handler.

Ex:
```javascript
// ==UserScript==
// @name     Testing APIs
// @version  1
// ==/UserScript==

function GM_notification(text, title, image, onclick) {
  let opt;

  if (typeof text == 'object') {
    opt = text;
    if (typeof title == 'function') opt.ondone = title;
  } else {
    opt = { title, text, image, onclick };
  }

  if (typeof opt.text != 'string') {
    throw new Error('GM.notification: "text" must be a string');
  }

  if (typeof opt.title != 'string') opt.title = 'Greasemonkey';
  if (typeof opt.image != 'string') opt.image = 'skin/icon32.png';

  let port = chrome.runtime.connect({name: 'UserScriptNotification'});
  port.onMessage.addListener(msg => {
    const msgType = msg.type;
    if (typeof opt[msgType] == 'function') opt[msgType]();
  });
  port.postMessage({
    name: 'create',
    details: {
        title: opt.title,
        text: opt.text,
        image: opt.image
    }
  });
}

GM_notification('Hello', 'Popsicle', null, function() {
  console.log('clicked');
});
```

arantius:
The above fix is packaged in version 4.2beta2:
https://addons.mozilla.org/firefox/downloads/file/833159/greasemonkey-4.2beta2-an+fx.xpi?src=devhub

Testing is always appreciated!

PR #2779, opened 2017-12-25
Add onMessage tests

Sxderp:
I was working on #2777 and came to the point where I should implement some tests. I found out that there was no setup for specifically testing `on-message`. As I worked to implement the testing bits I realized that more than minor setup / changes were needed in order to adequately provide the test environment and decided to work on that before going back the original issue.

Commit messages explain the changes pretty well I think.

Of note:
'Safely handle bad userscript save transactions' was cherry-picked into this branch.

### DIFF
--- a/doc/Messages.md
+++ b/doc/Messages.md
@@ -67,17 +67,6 @@ Data:
 * `uuid` The UUID of an installed script which is storing this value.
 * `value` The new value to store.
 
-# EditorSaved
-Sent by: `content/edit-user-script.js`.
-Received by: `bg/user-script-registry.js`.
-
-Sent whenever the user triggers the save action in the user script editor.
-Data:
-
-* `uuid` String UUID of the script being edited.
-* `content` String text content of main script.
-* `requires` Object mapping require URL to text content.
-
 # EnabledQuery
 Received by: `bg/is-enabled.js`.
 
@@ -117,25 +106,6 @@ progress as a percentage.  Sent specifically back to the content process
 
 * `errors` A (possibly empty) list of string error messages.
 * `progress` A number, 0.0 to 1.0, representing the completion so far.
-
-# UserScriptChanged
-Sent by: `bg/user-script-registry.js`
-
-Sent when some value (like enabled state) of a script is changed.  Data:
-
-* `details` Updated script's current.
-* `parsedDetails` Updated script's original parsed details.
-
-# UserScriptGet
-Sent by: `content/edit-user-script.js`
-
-Data:
-
-* `uuid` The UUID of an installed script to fetch.
-
-Response:
-
-* `details` The details object from an `EditableUserScript`.
 
 # UserScriptInstall
 Sent by: `content/install-dialog.js`
--- a/doc/Messages.md
+++ b/doc/Messages.md
@@ -78,13 +78,6 @@ Data:
 * `content` String text content of main script.
 * `requires` Object mapping require URL to text content.
 
-# EnabledChanged
-Sent by: `bg/is-enabled.js`.
-
-Sent whenever the global enabled status changes.
-
-* `enabled` boolean, the new status (true = enabled, false = disabled).
-
 # EnabledQuery
 Received by: `bg/is-enabled.js`.
 
--- a/doc/Ports.md
+++ b/doc/Ports.md
@@ -1,0 +1,50 @@
+This document keeps track of all the parent/child ports that Greasemonkey
+creates. All sections should look like:
+
+    # PortName
+    Sent by: `creator.js`
+    Received by: `listener.js`
+
+    Description of the purpose of this port, its important details, incoming
+    and outgoing message schema and expected responses. Message names are not
+    necessarily included when sending message but are intended to be used as
+    a labeling mechanism for the documentation.
+
+    An additional note on identifying message direction. Incoming messages are
+    sent from `creator.js` to `listener.js`. While outgoing messages are the
+    reverse.
+
+All message should specify `JSON.parse()`-able bodies, and always with a `name`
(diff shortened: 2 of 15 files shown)


arantius:
This PR is several completely separate things (namespace for message-related functions, documentation structure changes, change editor from messages to ports, something about transactions ... then finally what the description actually says?).  This makes it hard to review both the actual code, and the merit of each of these changes.

Please make one PR per logical change.

Sxderp:
> This PR is several completely separate things ...

Ah, I can see why it may look like they're all separate just looking at the commits. However, the only changes that I'd considered 'separate' would be the move of all `on*` functions to a namespace object. The reason this was done was to reduce pollution on the global object (window) after the move of `onMessage` to the global object (so that it's available in testing). Moving of those functions is not necessarily required and could be taken out if request. I still think the semantics look nice, but that's neither here nor there.

Anyway, onto the commits themselves.

First one, as mentioned `onMessage` was move to the global scope. This allows for the creation of plumbing for `chrome.sendMessage` in the testing environment ([setup.js](https://github.com/Sxderp/greasemonkey/blob/2c57cecd02462619f8dd65788f99666729730f53/test/setup.js)).

That's pretty good, since it will allow for better integration during tests, that is, due to the structure of Greasemonkey almost all messages are first handled by `onMessage` before being passed to their respective handlers. Therefore simulating `chrome.sendMessage` is basically just passing the message to `onMessage`.

This would have been great if that's all that was necessary. Unfortunately, since `chrome.sendMessage` was no longer an empty void in the testing environment, errors started popping up. Mainly printed ones. But the big item was an error that is almost entirely ignored in the current codebase (and really shouldn't be), [unhandled message error](https://github.com/Sxderp/greasemonkey/blob/2c57cecd02462619f8dd65788f99666729730f53/src/bg/on-message.js#L13).

Forward, leads us to the next commit, removing 'EnableChange'. This was one of those messages that did not have a handler. Therefore would spit an error out. Since there's no reason to keep dead code around (and I haven't seen any plans to use the message), it was removed. As the commit message mentioned, this caused those particular error messages to stop appearing in the tests.

This leaves us with the last 'out there' commit. As mentioned before, the often ignored 'unhandled message' error kept popping up. And it probably could have been removed or exceptions made or just ignored during the tests (which really just throws the problem under the bus). The last message that appeared as unhandled was 'UserScriptChanged'. Now, aside from the above methods, it's not fully obvious how this particular error should be resolved. Why? Because the 'UserScriptChange' is _indeed_ handled, just not by the background, so it's a false-positive error. The solution? Ports.

Aside from ports being _the_ interface for continuous communication between the content and background (which the editor needs), they also solve some other problems that messages had (and of course circumvent the 'unhandled message' error). The biggest benefit is that a port doesn't have to worry about accidentally intercepting messages intended for others. They receive messages only directed to them. This was a problem previously and, although fixed, could happen again.

---

The one commit that I glossed over was the 'safely handle bad transactions' from #2750. The reason is that I can't quite remember why I needed it. But there was some error that happened when migrating to using a ports interface that was resolved with that commit. Therefore it was cherry picked.

> documentation structure changes ...

I wouldn't really call it that. The documentation changes involved removing messages that were no longer being used (standard, as would be adding new messages). And adding a new file to describe the port interface created for the editor. I don't see how the new `Ports.md` could be separated. It _directly_ correspond to describing the interface between the editor and the background. If the style or format doesn't work for you then that's easily changed, but I think it's important to document those changes.

---

Now, could the commits be separated? Perhaps, but they are not completely unrelated commits. The biggest that I can agree could / should be separated is the namespacing.

The others could probably fall under the umbrella of 'fixing message handling errors that have been ignored' and be a separate pull request that is applied before adding the `chrome.sendMessage` plumbing to the test suite. But since the errors became prominent _after_ that change I figured they should just be resolved right away.

arantius:
> ... an error that is almost entirely ignored in the current codebase (and really shouldn't be), unhandled message error.

I hope that moving back to `browser` will make this better (because it's exceptions that can't go unhandled in that way).  I gave my first weak shot at starting that today, but in tests "ReferenceError: browser is not defined" and I'm unclear why.  The sinon-chrome tool claims that it provides both?  The NPM package provides one but not the other?

> The last message that appeared as unhandled was 'UserScriptChanged'. ... Because the 'UserScriptChange' is indeed handled, just not by the background, so it's a false-positive error. The solution? Ports.

Why isn't making the real and test environments match better (without structural change) not the solution?

> Aside from ports being the interface for continuous communication between the content and background (which the editor needs)

Why does the editor need this?

> The biggest benefit is that a port doesn't have to worry about accidentally intercepting messages intended for others.

Yes it does, that's why you have to switch on the name when you set up the listener.  (Once that is done, yes, then.  But you still have to do it right.)

Sxderp:
> Why isn't making the real and test environments match better (without structural change) not the solution?

It could be, but I figured using ports would both solve the problem and create a better, and isolated, interface for the editor to communicate with the background.

> Why does the editor need this?

Perhaps 'needs' is the incorrect word. However, there is still continuous messaging back and forth between the editor and the background. I would think that using a port, which is direct, would be more efficient than having to broadcast messages to everyone each time the editor saves, and the script is changed, no? I have no detailed performance stats but it seems reasonable.

> Yes it does, that's why you have to switch on the name when you set up the listener. (Once that is done, yes, then. But you still have to do it right.)

I meant on the editor side. If you have two editors opened to the same script you can have two different working copies. Say, you have some code and then have an idea, but don't know if it will work. You could open a new tab and then start making modifications (leaving the existing one open). With the messaging after you saved and got hit with the 'UserScriptChanged' the original tab would potentially have some of its data overwritten. The port would keep each one isolated. Sure, there's other ways one could go about doing the above, but it is still valid.

Issue #2693, opened 2017-11-17
User scripts for local files ("file:///" protocol)


FeBe95:
I know it is stated in the Greasemonkey (4) wiki that "User scripts will not run on documents from any other scheme (ftp, file, etc.) or any other part of about. ", but why is this the case? WebExtensions are able to access local files and Violentmonkey for example is able to execute user scripts on "file:///" pages.

StigNygaard:
#2612

FeBe95:
The issue you linked is about installing a local `myscript.user.js` file, whereas my issue is about executing user scripts on pages that are local files.
For example a script that redirects local php files `"file:://C:/xampp/....../index.php"` to `http://localhost/....../index.php`.

Sxderp:
I'm actually not sure about this. Executing a user script on a `file://` protocol shouldn't have any technical limitations.

arantius:
> Executing a user script on a file:// protocol shouldn't have any technical limitations.

Conceptually, no.  Today we `tabs.executeScript()`, which is just a content script, which is limited.  Mozilla really rushed their WebExtension decision.

I hope #2549 will help with this, but we have to wait for Firefox to unblock us there.

FeBe95:
Okay, but without Firefox changing something, Violentmonkey already is able to execute scripts on local pages. Are they doing something fundamentally different ?

Sxderp:
> Conceptually, no. Today we tabs.executeScript(), which is just a content script, which is limited. Mozilla really rushed their WebExtension decision.

Content scripts will run on `file://` There's just some edge case functionality in regard to certain background listeners [matching on them](https://github.com/greasemonkey/greasemonkey/issues/2612#issuecomment-340314062). And of course the async request for them which spawned that particular issue[1]. In fact `webNavigation.onCommitted` matches `file://` protocol just fine.

In the _current_ codebase the only thing preventing scripts from running on `file://` is [here](https://github.com/greasemonkey/greasemonkey/blob/64f1ae1945c54ad343a2aad896b72a71f86fd07a/src/user-script-obj.js#L136-L142).

[1] @erosman made some good points in regard to local file install workflow. I think it should be considered.

erosman:
> In the current codebase the only thing preventing scripts from running on file:// is here.
Is there an update to this?

Sxderp:
> Is there an update to this?

[...]

Issue #2652, opened 2017-11-09
Active/disabled flag not saved


McBen:
Greasemonkey 4.1beta2
Firefox 58.0b1 - Linux/Kubuntu

on FF startup GM is always active even if disable in previous session

Sxderp:
After tracing the toggle button through a message and quite a few function calls, we're eventually lead [here](https://github.com/greasemonkey/greasemonkey/blob/c194621bdc89cebbb8bd54ebef807fe5e4cdb9ae/src/bg/is-enabled.js#L26-L29). In which an 'EnabledChanged' message is sent but nothing is listening. So, the setting isn't saved nor is it actually loaded it's currently, from what I can tell, [hard coded as enabled](https://github.com/greasemonkey/greasemonkey/blob/c194621bdc89cebbb8bd54ebef807fe5e4cdb9ae/src/bg/is-enabled.js#L9) on startup.

Looks like this needs to be added as a database entry somewhere so it's saved.

arantius:
Indeed; `setGlobalEnabled()` should be doing some persistence.

The `EnabledChanged` message was an analog from 3.x whereby there were multiple places that it could be changed, so the change was broadcasted to enable all the UIs to update when any one of them was used.  Might be obsolete now.

arantius:
This shouldn't have been closed!

arantius:
This (user scripts execute at full extension-content-script privs) should be temporary, but for now I have to agree.

arantius:
We're already using local storage in one place.  And there's not much a script could do with access to this single value.  So for now, it's easy and we're going there.

arantius:
The above fix has been packaged in a new beta version:
https://addons.mozilla.org/firefox/downloads/file/830369/greasemonkey-4.2beta1-an+fx.xpi?src=devhub

Testing and confirmation would be appreciated!

arantius:
As written this causes Greasemonkey to disable itself upon first launch (because the setting is "enabled", is not present, so is not truthy).  That shouldn't happen.

arantius:
The above fix is packaged in version 4.2beta2:
https://addons.mozilla.org/firefox/downloads/file/833159/greasemonkey-4.2beta2-an+fx.xpi?src=devhub

Testing is always appreciated!

arantius:
Actually I mean b994449c94d71eaf5d33a245043dae4054774bf1 (where I typoed 2562 instead of 2652).

PR #2780, opened 2017-12-27
Allow scripts to execute on the `file:` protocol

Sxderp:
#2693

### DIFF
--- a/src/user-script-obj.js
+++ b/src/user-script-obj.js
@@ -149,6 +149,7 @@ window.RemoteUserScript = class RemoteUserScript {
     if (url
         && url.protocol != 'http:'
         && url.protocol != 'https:'
+        && url.protocol != 'file:'
         && !url.href.startsWith('about:blank')
     ) {
       return false;
--- a/test/user-script-obj.test.js
+++ b/test/user-script-obj.test.js
@@ -28,25 +28,49 @@ gt_one(2);
   });
 
   describe('RemoteUserScript.runsAt()', () => {
-    let userScript = new RemoteUserScript({});
-    let url = new URL('http://example.org/');
+    let userScript;
+    const matches = urlStr => assert.isOk(userScript.runsAt(new URL(urlStr)));
+    const notMatches =
+        urlStr => assert.isNotOk(userScript.runsAt(new URL(urlStr)));
 
-    it('fails gracefully with a non-MatchPattern object', () => {
-      userScript._matches = [{}];
-      let result = userScript.runsAt(url);
-      assert.equal(result, false);
+    beforeEach(() => {
+      userScript = new RemoteUserScript({});
     });
 
-    it('works with MatchPattern object', () => {
-      userScript._matches = [new MatchPattern('http://*/*')];
-      let result = userScript.runsAt(url);
-      assert.equal(result, true);
+    describe('basic functionality', () => {
+      const url = 'http://example.org/';
+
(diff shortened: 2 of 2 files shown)


Sxderp:
In #2782 I mentioned that I didn't recall how I stumbled upon the error. Well, I figured it out. It was when I was working on this pull request. I found that out as I was going over the tests that I had `git stash` (and subsequently forgot about).

I've rebased this pull on to #2782. As well as pushed the tests.

arantius:
I checked out this branch and installed this script:

```
// ==UserScript==
// @name     Run-on-file
// @version  1
// @grant    none
// @include  file:///*
// ==/UserScript==

console.log('GM script running on file!');
```

But I don't see the log message.  This switches to `webNavigation.onCommitted` to detect execution time, which AFAICT does not fire for files (because they're not "web" navigations?).

Sxderp:
> But I don't see the log message. 

Interesting. At the moment I don't have access to newer versions of FF to test, but it worked on RHEL6 FF 52.5.1. And if I'm remembering correctly it should have worked on elementaryOS (Ubuntu based) FF 57.? (although it _may_ have been on nightly and therefore FF 59.?).

I don't recall testing on my Windows.

arantius:
I'm testing on Ubuntu, FF 59.

erosman:
> But I don't see the log message. This switches to webNavigation.onCommitted to detect execution time, which AFAICT does not fire for files (because they're not "web" navigations?).

I have an addon where I use `webRequest` which also doesn't fire on local locations i.e. `about: view-source: moz-extension: chrome: resource: file:`
I guess they used the same approach for `webNavigation`

I overcame the problem by adding `tabs.onUpdated` listener.

Note: `tabs.onUpdated` fires too many times.
You would need to check for `changeInfo.url` and then use `tabInfo.url` and then avoid reinjecting the same script again.

However ............... It says here as well the API doesn't allow injection to file at all (unless manually allowed by the user) :(

https://stackoverflow.com/questions/33575973/how-to-get-content-scripts-to-execute-in-file-urls-in-a-chrome-extension

https://stackoverflow.com/questions/34250957/google-chrome-chromium-event-filter-local-file-hostname

Sxderp:
> I'm testing on Ubuntu, FF 59.

Weird. I just double checked on elementaryOS FF 57 + 59, both worked. Also tested on Windows 59, worked as well.

I could start up a Ubuntu VM and see if it's just that platform (which is again, weird).

---

I tested by typing the path into the URL and by running `firefox ./test.html`. I was unable to test drag + drop on the elementaryOS (RIP, can't seem to drag files using whatever the standard manager is). But drag + drop worked on Windows.

Sxderp:
Oh, I just rebased this and #2782 on master. So if you test, then make sure you account for #2652

erosman:
Is there a beta to test?

Sxderp:
> Is there a beta to test?

Nothing has been pushed to AMO. You can always check out my branch and manually install from file or load temporary add-on.

Sxderp:
I grabbed an Ubuntu VM (nothing to install an actual copy on at the moment) and see no problems. After I enabled GM (since it's currently off by default), I performed the same tests as above, drag+drop, URI navigation, and `firefox index.html`. With all three I saw the logged message in the page console.

arantius:
Tried to test again and I agree that it works.  Wonder what I did wrong before.

PR #2782, opened 2017-12-28
Fix matchpattern `file:` match

Sxderp:
I don't recall how I stumbled upon this, but I found out that `MatchPattern.js` doesn't quite follow Google's specification. `file:///foo*` is a valid match pattern but the previous code would not allow it. A fix to the Regex solved that problem.

Also added a bunch of tests.

### DIFF
--- a/test/third-party/MatchPattern.test.js
+++ b/test/third-party/MatchPattern.test.js
@@ -1,48 +1,47 @@
 describe('third-party/MatchPattern', () => {
     describe('invalid patterns', () => {
-      const newPattern =
-        pattern => { return () => new MatchPattern(pattern) };
+      const newPattern = pattern => { return () => new MatchPattern(pattern) };
 
       it('http:///foo*bar', () => {
         // No host provided
         expect(newPattern('http:///foo*bar'))
-          .to.throw('@match: No host specified for (http:).');
+            .to.throw('@match: No host specified for (http:).');
       });
 
       it('file://value.com/foo*bar', () => {
           // Host improperly provided for file
           expect(newPattern('file://value.com/foo*bar'))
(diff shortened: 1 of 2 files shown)


Eselce:
TYPO: protcol

Sxderp:
> TYPO: protcol

Never happened. =-)

test/third-party/MatchPattern.test.js
```diff
@@ -1,10 +1,211 @@
 describe('third-party/MatchPattern', () => {
-  const matches
-      = (pattern, urlStr) => assert.isOk(pattern.doMatch(new URL(urlStr)));
-  const notMatches
-      = (pattern, urlStr) => assert.isNotOk(pattern.doMatch(new URL(urlStr)));
+    describe('invalid patterns', () => {
+      const newPattern =
+        pattern => { return () => new MatchPattern(pattern) };
```
arantius: I think this fits on one line, but for all statements continued across >1 line, continuations should be indented 4 spaces.

arantius:
This is valuable, but only after #2693 .

Sxderp:
> This is valuable, but only after #2693 .

It's the other way around. This PR, `fixing MatchPattern` #2782, needs to be merged before #2693 (or any PRs that fix it).

There's no point in fixing / allow #2693 if MatchPattern doesn't properly match against them.

arantius:
if only #2693 is done you can still use `@include`, and get execution of scripts at local files.

If you only do this, you get nothing.  But yes, the two are closely related.

Sxderp:
> if only #2693 is done you can still use @include

Heh, I forgot about `@include`. I'll rebase the commits.

Issue #2785, opened 2017-12-30
Add a SAVE Button to the Editor


erosman:
Adding a save button (well, tab like button next to the name tab in editor) would make it more user-friendly for the average user and it should be easy to implement.

arantius:
The above fix has been packaged in a new beta version:
https://addons.mozilla.org/firefox/downloads/file/830369/greasemonkey-4.2beta1-an+fx.xpi?src=devhub

Testing and confirmation would be appreciated!

erosman:
> Testing and confirmation would be appreciated!
I am going to try 4.2b and let you know

Eselce:
> #2693 doesnt work. How come it was removed?

It just doesn't work as expected, see the pull request thread...

PR #2788, opened 2018-01-05
Add a save button

badbrainz:
This adds a 'save' button next to the tabs. #2785


arantius:

PR #2789, opened 2018-01-06
Avoid crash in setIcon on Firefox for Android

ToadKing:
Trying to toggle greasemonkey on or off fails right now on Android due to a missing API. Check for setIcon before we call it.

https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/browserAction/setIcon#Browser_compatibility


arantius:
Thanks!  Android support is still not really there.  I think you can install but not manage scripts yet, but every little bit helps.

PR #2791, opened 2018-01-06
Repair keyboard navigation on user script list in monkey menu.

esperecyan:
Document#getElementsByTagName() returns live collection but ParentNode#querySelectorAll() returns static collection.


arantius:
TIL.  Never tried to use a live collection before, didn't realize it was a thing.

Issue #2771, opened 2017-12-14
xmlHttpRequest.open no longer accepts path without protocol and host


jesus2099:
II have noticed in jesus2099/konami-command#372 that Firefox v57 + Greasemonkey v4 would only work with full URL `https?://host.tld/pathname` and no longer works with just `/pathname`.
Firefox v57 + another GM engine is not affected by this regression, this is why I open a ticket here.

jesus2099:
In my [work‐around](https://github.com/jesus2099/konami-command/commit/c9d957184d365424967d1a2d57636a55d97e6a1b#diff-89146a92c880500983b57ea698ab9d27R213), I had to append `self.location.protocol + "//" + self.location.host + ` in front of my URL.

Sxderp:
Running requests from a content script context (which userscripts are run in) must use full urls.

https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Chrome_incompatibilities#Content_script_requests_happen_in_the_context_of_extension_not_content_page

If you don't want to use full urls then you can use the methods on `window.wrappedJSObject` and export any functions that are needed to run (onreadystatechange, etc) using `exportFunction()`. My suggestion is just to use full urls.

jesus2099:
Thanks very much for your help.
OK so you mean it’s per specifications of the WebExtensions system, right?
I don’t know this WebExtensions stuff, but it seems to be a standard…?
I don’t remember reading anything about this particular regression in the [GM v4 for authors](https://www.greasespot.net/2017/09/greasemonkey-4-for-script-authors.html) blog posts.
It’s sure this regression is permanent?

Sxderp:
> It’s sure this regression is permanent?

Maybe. I don't really know. Perhaps down the line `@grant none` will not have these issues, if a proper sandbox can be created. But if some privileged APIs are required then these steps may be necessary. The ball's really in Mozilla's court.

jesus2099:
FTR my script is a `@grant none` script.
I let the project leaders close or not close this ticket.
I’ll adapt.

arantius:
This is not intended, so it shouldn't be expected to be permanent.  We could totally resolve non-absolute paths based on the document as a root.

iology:
Please note that the correct base url is: `document.baseURI` (related to `<base href="...">`)

dogancelik:
I can confirm, AJAX doesn't work with only path URL

```
$.ajax({
  url: '/some/path',
  success: () => {}
});
[...]

PR #2792, opened 2018-01-07
fix XMLHttpRequests that don't have the "async" parameter

ToadKing:
The #2771 fix has a bug in it. If you don't supply the optional async parameter it defaults to undefined, which evalulates to false. It should default to true if it's not supplied. This fixes the xhr shim.


ToadKing:
Made some extra changes to the shim:

* Replaces XMLHttpRequest.prototype.open instead of wrapping the XMLHttpRequest constructor. Also fixes scripts that might use the prototype directly bypassing the fix.
* Shim block is no longer in a function. Not needed anymore since we don't use `var` anywhere and don't have to worry about hoisting.

PR #2793, opened 2018-01-07
Support downgrading evalContentVersion on scripts

ToadKing:
Ran into this developing my last patch. Allowing for version downgrades on evalContent should be allowed.

### DIFF
--- a/src/bg/user-script-registry.js
+++ b/src/bg/user-script-registry.js
@@ -106,7 +106,7 @@ function loadUserScripts() {
       event.target.result.forEach(details => {
         let userScript = new EditableUserScript(details);
         userScripts[details.uuid] = userScript;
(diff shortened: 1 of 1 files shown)


Sxderp:
Agreed, I've run into this a few times before too. I'd just remove the script and move on. But this is better.

arantius:
Can you clarify this for me?  What's the expected situation where this number goes down?

Sxderp:
> Can you clarify this for me? What's the expected situation where this number goes down?

Mostly development builds. Suppose your testing something and then roll back. Sometimes you're stuck with incompatible scripts and need make a new one, copy paste, and continue on.

ToadKing:
It can also be useful to users downgrading from a beta release to a stable release, in case the evalContent in the beta breaks something.

PR #2798, opened 2018-01-11
Make `UserScriptRegistry._loadUserScripts()` return Promise for tests.

esperecyan:
- `indexedDB.deleteDatabase('greasemonkey')` will be blocked until tests completed because user-script-registry.js leave the database openning.
- Now the second `assert.isOk(scriptNamed('footnote'));` in user-script-registry.test.js will be executed before executing `UserScriptRegistry._loadUserScripts();` is completed.

### DIFF
--- a/src/bg/user-script-registry.js
+++ b/src/bg/user-script-registry.js
@@ -130,25 +130,29 @@ async function loadUserScripts() {
 
 async function loadUserScripts() {
   let db = await openDb();
-  let txn = db.transaction([scriptStoreName], "readonly");
-  let store = txn.objectStore(scriptStoreName);
-  let req = store.getAll();
-  req.onsuccess = event => {
-    userScripts = {};
-    event.target.result.forEach(details => {
-      let userScript = new EditableUserScript(details);
-      userScripts[details.uuid] = userScript;
-      if (userScript.evalContentVersion != EVAL_CONTENT_VERSION) {
-        userScript.calculateEvalContent();
-        saveUserScript(userScript);
-      }
-    });
-    db.close();
-  };
-  req.onerror = event => {
-    console.error('loadUserScripts() failure', event);
-    db.close();
-  };
+  return new Promise((resolve, reject) => {
+    let txn = db.transaction([scriptStoreName], "readonly");
+    let store = txn.objectStore(scriptStoreName);
+    let req = store.getAll();
+    req.onsuccess = async event => {
+      userScripts = {};
+      await Promise.all(event.target.result.map(async details => {
+        let userScript = new EditableUserScript(details);
+        userScripts[details.uuid] = userScript;
+        if (userScript.evalContentVersion != EVAL_CONTENT_VERSION) {
+          userScript.calculateEvalContent();
+          await saveUserScript(userScript);
+        }
+      }));
+      resolve(); 
+      db.close();
+    };
+    req.onerror = event => {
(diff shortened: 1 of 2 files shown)


src/bg/user-script-registry.js
```diff
@@ -97,28 +97,49 @@ async function installFromSource(source) {
...
+      req.onerror = event => {
+        console.error('loadUserScripts() failure', event);
+        reject(event.target.error);
+      };
+    });
   });
 };
 
 
+function clearObjectStore() {
```
Sxderp: This doesn't clear `value-store` databases. At the moment this isn't needed in testing, nor the actual code, but it would be useful for both the testing and the code[1]. I suggest creating a function to delete individual databases in `value-store.js` and then looping through `userScripts` creating a list of Promises that could be `Promise#all()`.

In my branch for [1] I just updated `UserRegistry#uninstall` to include the `value-store.js` cleanup, then to remove all scripts I looped through `userScripts` calling the uninstall function. Somewhat inefficient but ensured that all cleanup code was executed. Which may be important should additional cleanup code be required down the line.

[1] For example import / exporting the database information. Clearing the database before import was an option that I provided.

test/bg/user-script-registry.test.js
```diff
@@ -1,6 +1,6 @@
 describe('bg/user-script-registry', () => {
-  beforeEach(() => {
-    indexedDB.deleteDatabase('greasemonkey');
+  beforeEach(async () => {
+    await UserScriptRegistry._clearObjectStore();
```
Sxderp: I believe this could be simplified to just `UserScriptRegistry._clearObjectStore` (since it now returns a promise).

src/bg/user-script-registry.js
esperecyan: Yes, this Pull Request is just for fixing `indexedDB.deleteDatabase('greasemonkey');` that is not working now.
I think that to replace with complete cleanup code is more preferred too.
Should I remove this commit and wait for your branch to be merged?

test/bg/user-script-registry.test.js
esperecyan: I wrote this as one part of beforeEach code but is it better to write `beforeEach(UserScriptRegistry._clearObjectStore);` better for code style?

esperecyan:
I modified the commits into one which just deletes `indexedDB.deleteDatabase('greasemonkey')`.

arantius:
Tests shouldn't leave side effects around.  If they're going to be writing to indexedDB, they should leave things clean afterwards.  Looks like you've figured out that the attempt to do this isn't working, but you should fix it rather than just remove it if so.

esperecyan:
I narrowed this PR down to fixing `UserScriptRegistry._loadUserScripts()` because resolving the issue about `deleteDatabase` seems to be proceeding in https://github.com/greasemonkey/greasemonkey/compare/master...Sxderp:import-export-database-merge .

Sxderp:
> seems to be proceeding in master...Sxderp:import-export-database-merge .

Which is waiting on a few things, namely #2801 , some more testing on my part, and probably a few more unit tests. As well as a bit of code cleanup.

esperecyan:
I resolved conflicts with https://github.com/greasemonkey/greasemonkey/commit/74bc5dcd3e533384e6adaa1f9961e4fbdfcfa427 .

Issue #2747, opened 2017-12-05
Add backup/restore


Rick-74:
With old greasemonkey an export scripts/preferences function wasn't really needed, since the user could just move the gm_scripts directory from one profile/computer to another: but now this is no longer possible, so please consider adding such a function to greasemonkey 4 in the future.

Sxderp:
Well, Firefox and its issues. So, I was implementing this (and of course an 'import' function). I'm almost done with it, but there's a show-stopper. It only works when you open the `browser toolbox` and select "do not autoclose popups". The import is done through a `<input type="file">` (which is the only way), however, on focus of the browse window the popup normally closes, thus the window is destroyed and no functions are executed. Relevant bugs:
https://bugzilla.mozilla.org/show_bug.cgi?id=1292701
https://bugzilla.mozilla.org/show_bug.cgi?id=1366330

It's so annoying having a half-assed environment to work in.

Also, applies to #2612. Which can be addressed at the same time this issue is.

makyen:
@Sxderp If the popup closing is an issue, the normal way to solve/work-around that is to just open a tab or window (pointing to an HTML page within the extension) in which you perform whatever you need to do? Such tabs/windows are not auto-closed and present the same environment as a popup. While that might not be as clean as having it all within the browser popup, it should be functional.

Sxderp:
I could do that. It'd involve moving a bunch of code around. I'll look into it some other time.

I've completed the import / export stuff. No tests, and very little error handling. But it works. Though, for simplicity I made it so that an import would overwrite the entire database. With a confirmation prompt of course.

arantius:
I think ideally the user could choose to replace or merge the import.  Stylus does only merge, which is also bad IME.  User choice, with defaults that might work, seems best.

Sxderp:
Sure, but then we run into a lot of conditionals. I'd like to see a flow diagram of some sort on the _exact_ nature of merging. 

Mostly in terms of conflicts. Overwrite content, overwrite key/values (yes those are exported too), prompt for everything (that seems like a bad user experience), etc?

arantius:
I mean just: import every script in the file (and completely overwrite to that state), but don't touch/delete scripts that aren't in the file.

Sxderp:
Oh. That's far simpler. Yeah, that sounds good.

Sxderp:
So I've implemented three different import methods.

[...]

PR #2801, opened 2018-01-13
Separate value-store implementation from messaging

Sxderp:
Unlink value-store functions from the messaging interface so that
they can be called directly from the background script, should the
need arise.

Useful for tests and for #2747


Sxderp:
Added tests and a way to delete a store (required to cleanup tests).

PR #2802, opened 2018-01-14
Make `_randomUuid()` always return string of length 36.

esperecyan:
`randomInts[i].toString(16)` returns a string of length 1 or 2.

### DIFF
--- a/test/user-script-obj.test.js
+++ b/test/user-script-obj.test.js
@@ -104,4 +104,65 @@ gt_one(2);
       });
     });
   });
+
+  describe('RunnableUserScript.uuid', () => {
+    describe('is set into a valid version 4 UUID', () => {
+      afterEach(() => {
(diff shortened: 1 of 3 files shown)


test/user-script-obj.test.js
```diff
@@ -104,4 +104,15 @@ gt_one(2);
       });
     });
   });
+
+  describe('RunnableUserScript.uuid', () => {
+    it('is set into a valid version 4 UUID', () => {
+      for (let i = 0; i < 1000; i++) {
```
arantius: This is not a good test.  It's wasteful, and not guaranteed to check the condition we want.  Maybe stub (or dependency inject) `getRandomValues` to return at least one value below 16, and instead assert that it's formatted correctly?

esperecyan: OK. I changed it.

Issue #2612, opened 2017-10-12
Allow installing local files


arantius:
In the past you could File>Open a `.user.js` and it would install.  In 4.0 so far, it does nothing, just opens the file.

Sxderp:
Perhaps [this](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Match_patterns) (match pattern docs)? Using `*` in the [matches selector](https://github.com/greasemonkey/greasemonkey/blob/029d1e595c0f0e1a4e10acbe95b186be9690d5ae/manifest.json#L46) for the scheme only globs to http or https.

ToadKing:
That or the `<all_urls>` selector.

arantius:
Adding a match pattern to detect `file:///` scripts just causes us to start an XHR which fails to read the contents at that URL, which is surprising.

Sxderp:
I've been looking into this a bit. And I've come to two conclusions. The issue may have to do with disallowing filesystem access to WebExtensions, even if readonly through XHR (no source on this directly). Or it could be related to [same origin policy](https://developer.mozilla.org/en-US/docs/Same-origin_policy_for_file:_URIs).
> Starting in Gecko 1.9, files are allowed to read only certain other files.  Specifically, a file can read another file only if the parent directory of the originating file is an ancestor directory of the target file.

arantius:
Yeah, it's probably the "no files" rule.  Don't know where to get official guidance on that but I know it's true now that I think about it.  You only get very special exceptions like the downloads API for making new files.

We could try to figure a workaround like directly handling the drag/drop, or (LOL) reading the content of the tab that's already open.  But then we'd fail to fetch relative icon/resource/requires, so it still won't go, without even more workarounds.

Sxderp:
Could you not use storage.local to cache the content using the url as a key before navigating the to install page? You already have the content in a variable. Of course the cache would have to be cleared once installed / decidedly not installed. Would be more convenient if WebExtensions had some sort of temporary storage so they don't have to manually deal with eviction.

Sxderp:
I've been messing around with what I've said and it's not as straightforward as I would have thought. Firstly, any user script has access to `browser.storage.local` so it's inherently an unsafe storage for addons like Greasemonkey. Secondly, the same goes for sending the content through a message to the background script. I'm not really sure how to secure that to ensure the message was sent only from `script-detect.js`.

Sxderp:
[...]

PR #2804, opened 2018-01-14
Add import userscript from file

Sxderp:
Addresses #2612.

Not as nice as being able to navigate to the file but the the baseline result is the same. Unfortunately due to Mozilla issues with the panel a new tab has to be opened. When the relevant bug is fixed the import code should be copy/pastable directly into the Monkey Menu (necessary elements are in place).

I also have another branch that deals with import/export of the entire database, but that's somewhat reliant on the value-store PR (well, it could be done without the PR, but it makes things a bit simpler).

### DIFF
--- a/src/browser/monkey-menu.js
+++ b/src/browser/monkey-menu.js
@@ -112,6 +112,12 @@ function onHashChange(event) {
         focusSelection();
       }
       break;
+
+    // Imports are handled by onchange listeners or on keypress
+    case '#import-user-script':
+      history.replaceState({}, 'Home', '#menu-top');
+      break;
+
     default:
       // Check if it's a Userscript by examing the gUserScript object
       let userScript = gUserScripts[hash.slice(1)];
@@ -153,10 +159,26 @@ function onHashChange(event) {
 }
 
 
+// Import a new user script
+// TODO: Add onImportUserScript to monkey menu when file picker bugs are fixed
+// https://bugzilla.mozilla.org/show_bug.cgi?id=1366330
+function onImportUserScript(event) {
+}
+
+
 function onKeypress(event) {
   let key = event.key;
   if ('Enter' === key) {
     gLastHashChangeWasKey = true;
+    // Unfortunately file inputs cannot be opened in the hashchange because
+    // it's not considered a 'user action' context
+    if (event.target.classList.contains('input-item')) {
+      event.preventDefault();
+      // Create a non-bubbling event.
+      let fakeEvent =
+          new MouseEvent('click', {'bubbles': false, 'cancelable': true});
+      event.target.getElementsByTagName('input')[0].dispatchEvent(fakeEvent);
+    }
     return;
   }
   event.preventDefault();
@@ -207,6 +229,21 @@ function onLoad(event) {
       });
     }
   }
+
+  // TODO: Remove when file picker bugs are fixed
+  // https://bugzilla.mozilla.org/show_bug.cgi?id=1366330
+  // document.getElementById('import-script-file')
+  //     .addEventListener('change', onImportUserScript);
+  function preventAll(event) {
+    event.preventDefault();
+    event.stopPropagation();
+    event.stopImmediatePropagation();
+  }
+  document.getElementById('import-script-file')
+      .addEventListener('click', event => {
+        preventAll(event);
+        openTemporaryImportPage('import-script.html');
+      });
 }
 
 
@@ -214,6 +251,17 @@ function onUnload(event) {
   // Clear the pending uninstall ticker and cleanup any pending installs.
   clearInterval(gPendingTicker);
   checkPendingUninstall();
+}
+
+
+// TODO: Remove when file picker bugs are fixed
+// https://bugzilla.mozilla.org/show_bug.cgi?id=1366330
+function openTemporaryImportPage(page) {
+  chrome.tabs.create({
+    'active': true,
+    'url': chrome.runtime.getURL('src/content/import/' + page),
+  });
+  window.close();
 }
 
 
--- a/src/parse-user-script.js
+++ b/src/parse-user-script.js
@@ -35,6 +35,47 @@ function safeURL(path, base) {
 }
 
 
+function parseRemoteResources(details) {
+  let currentUrl;
+  let remoteType;
+  let downloadUrl = details.downloadUrl;
+
+  try {
+    if (details.iconUrl) {
+      remoteType = 'icon'
+      currentUrl = details.iconUrl;
+      details.iconUrl = safeURL(currentUrl, downloadUrl).toString();
+    }
+
+    let requireUrls = details.requireUrls;
+    if (requireUrls.length) {
+      remoteType = 'require';
+      for (let idx = requireUrls.length; idx--;) {
+        currentUrl = requireUrls[idx];
+        requireUrls[idx] = safeURL(currentUrl, downloadUrl).toString();
+      }
+    }
+
+    let resourceKeys = Object.keys(details.resourceUrls);
+    if (resourceKeys.length) {
+      let resourceUrls = details.resourceUrls;
+      remoteType = 'resource';
+      for (key of resourceKeys) {
+        currentUrl = resourceUrls[key];
+        resourceUrls[key] = safeURL(currentUrl, downloadUrl).toString();
+      }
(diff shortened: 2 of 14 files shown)


arantius:
Extract this zip file: [icon.zip](https://github.com/greasemonkey/greasemonkey/files/1668267/icon.zip)

Pick import, browse to that `.user.js`.  Console:

    at installFromSource(), db fail: TypeError: alert.png is not a valid URL.
    Stack trace:
    safeURL@moz-extension://ec66ea3f-5475-415e-b7af-7f37a4899dfe/src/parse-user-script.js:33:12
    window.parseUserScript@moz-extension://ec66ea3f-5475-415e-b7af-7f37a4899dfe/src/parse-user-script.js:124:25
    installFromSource/</<@moz-extension://ec66ea3f-5475-415e-b7af-7f37a4899dfe/src/bg/user-script-registry.js:72:23
    ...

After a couple seconds, a broken editor with no script pops up.  Nothing was installed.  I believe this will happen for any relative resource reference, this one is an `@icon`, but `@resource` or `@require` would do it too.

Sxderp:
> After a couple seconds, a broken editor with no script pops up. Nothing was installed. I believe this will happen for any relative resource reference, this one is an `@icon`, but `@resource` or `@require` would do it too.

How would a WebExtension have access to these files in order to import them? I suppose you could import the entire `.zip`. But this was just for a single userscript. And of course, if you have relative paths (no scheme) in a remote script, you're going to hit errors too.

Sxderp:
Alternatively, are you just pointing out the point that the editor pops open and shouldn't? That is fixable. As for actually getting that script to install, I'm not sure.

arantius:
> And of course, if you have relative paths (no scheme) in a remote script, you're going to hit errors too.

Try e.g. https://arantius.com/misc/greasemonkey/protect-textarea.user.js which contains the line `// @icon        img/textarea-padlock.png` and works perfectly.

> Alternatively, are you just pointing out the point that the editor pops open and shouldn't?

More generally that the error condition is not handled (because it's not even detected?).  And predicting the users that will see exactly the feature they want, except it doesn't work in the case they want to use it for.

Sxderp:
> Try e.g. ..

Ah, okay. I must have missed that functionality. Cool.

> More generally that the error condition is not handled (because it's not even detected?).

Detecting the error is doable. Of course getting the resources from the disk is probably a no-go. However, getting them remotely may be possible. Of course the implicit downloadUrl set due to navigation / remote install won't be applied from a local install, but it might be possible to modify  `parse-user-script.js` to use any explicit `@downloadUrl` metatag when resolving resource locations.

Sxderp:
Added an error if the URL cannot be resolved to a full path (with notification). If  `@downloadURL` is present then that will be used as the base. Of course, the current installation flow for `installFromSource` does not download remote data. So nothing is fetched. However, that can be resolved in a later PR I think (I'm working on one, but it still needs work).

src/browser/monkey-menu.js
```diff
@@ -207,6 +229,21 @@ function onLoad(event) {
...
+  // TODO: Remove when file picker bugs are fixed
+  // https://bugzilla.mozilla.org/show_bug.cgi?id=1366330
+  // document.getElementById('import-script-file')
+  //     .addEventListener('change', onImportUserScript);
+  function preventAll(event) {
+    event.preventDefault();
+    event.stopPropagation();
+    event.stopImmediatePropagation();
+  }
+  document.getElementById('import-script-file')
```
arantius: The element with this ID is hidden by CSS, so is not clickable, so this code doesn't work.

src/parse-user-script.js
```diff
@@ -77,6 +118,9 @@ window.parseUserScript = function(content, url, failIfMissing) {
     case 'noframes':
       details.noFrames = true;
       break;
+    case 'downloadURL':
```
arantius: It was a very intentional decision to not handle downloadURL, updateURL, etc. in GM4.  I won't merge code that tries to add them again.

This causes a situation that breaks the principle of least astonishment.  I might choose to save a script to my own disk/server, perhaps to isolate it from the original author.  But then by accident through one little setting it downloads/updates to some other remote resource that someone else controls?  Do not want.

src/browser/monkey-menu.js
```diff
@@ -153,10 +159,26 @@ function onHashChange(event) {
 }
 
 
+// Import a new user script
+// TODO: Add onImportUserScript to monkey menu when file picker bugs are fixed
```
arantius: I find it counterproductive and confusing to have this whole thing written half in one style and half in another.  I'm also not keen on the awkward wording in the UI.

Write it so that it works well, now, complete.  If and only if in the (distant, if ever) future something different becomes possible, then and only then write code to do that.

src/browser/monkey-menu.js
Sxderp: It is indeed clickable. Otherwise the installation page would never open.

src/parse-user-script.js
Sxderp: What do you propose then? We can't get the resources from disk without the user explicitly asking to open a file. As far as I'm aware we can't even open a 'please select next file' picker. Even if we could, asking the user to select every resource is pretty bad UX. Granted, the resources aren't downloaded due to current limitations of `installFromSource` but that should eventually be fixed.

So, should the script be imported without the external resources? (Yes, that's what happens now, but at least with the current PR the URLs are available and can be used to fetch resources).

Everything could be `.zip` but that could be pretty extreme for just a single script.

Alternatively, I could make it so that the original request takes precedence over the `@downloadURL`, that way if you're hosting it somewhere then resources will be fetched from your host, but if you're installing it by just importing the script it can still fetch remote resources using `@downloadURL`.

Or do we just not support relative paths? Which will probably break a lot of compatibility, no?

Sxderp: I should also say that some of the above also applies to the full DB import / export. Of course if you're importing an archive created by GM then you have access to everything as it was. However, if you import an archive exported from TM or VM then the only 'bits' that GM understands is the raw script file itself and the value-store (if importing from TM). The previously set, internal, `downloadUrl` will be unavailable (non-existant) and we're stuck with not being able to fetch resources.

src/browser/monkey-menu.js
arantius: When I check out this branch I see:

untitled

The \<a href="#import-user-script" ...\> and its text, but no \<input\>.  Clicking the (visible) anchor does nothing.  The CSS sets opacity to zero so maybe it's supposed to be overlaid (towards what end?), but that's broken.  If remove the opacity setting:

untitled1

The position is off?  Like I said above, the half-this-half-that implementation is suboptimal.

src/parse-user-script.js
arantius: Not all scripts specify `@downloadURL`, so users would be left with sometimes working sometimes not, even with this.

It would make plenty sense to me to allow: 1) `.user.js` import, if no deps (or all fully-specified and currently downloadable remote refs) and also 2) `.zip` import where there's exactly one `.user.js` and relative-only references to deps otherwise in the zip.

There's also some way to select a directory with a file browser?  I don't know what access you get to files in that directory though.

src/browser/monkey-menu.js
Sxderp: Ah, must have to do with the recent fixes to the MM styling. When I tested it just happen to be in the right location to work.

> (towards what end?),

I thought this was obvious. You can't style input fields. So either you hide it and overlay it on text you would like to display or you're stuck with a "browse" button with absolutely no context. Now, I can grant you that at this moment in time the file picker is unnecessary, since they don't work (RIP Mozilla please). But down the line, unless you want to stick with the poor UI of a file input, the overlay is pretty much required.

Either way, I'll remove the boilerplate as it's, as you said, not needed at the time.

Sxderp:
I'm going to close this and open a new one, branching from the initial commit. The new PR will focus on `1)` from https://github.com/greasemonkey/greasemonkey/pull/2804#discussion_r165739803. Adding `2)` or `3)` I think can be done on a later date since they focus on what's probably more of an edge case.

I'm also going to work on my import / export DB branch to follow the same ruleset as above. If an importing script doesn't have fully qualified resource paths then it won't be imported (unless the archive contains the GM 'bits').

Issue #2675, opened 2017-11-15
Use the "persistent" IndexedDB option


kekkc:
Hi, 

after cleaning & deleting my "profile\storage\default" I realized that GM4.0 is currently saving the user scripts there. The usual extension storage "browser-extension-data\ID" just contains a storage.js with the content:
"{"previousWelcomeUrl":"https://www.greasespot.net/2017/09/greasemonkey-4-announcement.html"}"

Wouldn't it be better to use "browser-extension-data\ID" ffor all GM user extension data (incl. user scripts)? All my other extensions (adblock plus...) are using this folder, whereas "profile\storage\default" is usually just used by websites. This folder is a data security risk and will be automatically cleaned by FF:

https://www.heise.de/newsticker/meldung/Gefaehrdeter-Datenschutz-Firefox-loescht-lokale-Datenbanken-nicht-3835084.html
https://bugzilla.mozilla.org/show_bug.cgi?id=1047098

Greetz

arantius:
> This folder is a data security risk

[citation needed]

kekkc:
My fear is that my user scripts will be deleted if I clear my browsing history.

https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/storage 
Each extension has its own storage area, which can be split into different types of storage.

Although this API is similar to Window.localStorage it is **recommended that you don't use Window.localStorage in the extension code to store extension-related data**. **Firefox will clear data stored by extensions using the localStorage API in various scenarios** where users clear their browsing history and data for privacy reasons, while data saved using the storage.local API will be correctly persisted in these scenarios.

Sxderp:
I don't know the folders but I do know the API points.
Greasemonkey stores userscripts in an `indexedDB`, which I'm going to assume is located in `profile\storage\default` as you have pointed out. Firefox will not _intentionally_ remove WebExtension data stored in `indexedDB`, actually I'm pretty sure Firefox will not _intentionally_ remove any WebExtension data which uses any of the storage WebExtension APIs.

> Although this API is similar to Window.localStorage it is recommended that you don't use Window.localStorage

`indexedDB` and `storage` are different from `window.localStorage`. What the message is saying is to not use `window.localStorage`. Greasemonkey does not use it.

> This folder is a data security risk

[...]

PR #2812, opened 2018-01-18
Add persistence to GM databases

Sxderp:
See #2675 for discussions.

Persistent DBs will only be used for FF >= 57. The reason is that the persistent APIs are not supported on older versions. The only way to persist is to use the non-standard `{storage: 'persist'}` option. Unfortunately that option makes migration from 'non-persisted' to 'persisted' very difficult.

See [this comment](https://github.com/greasemonkey/greasemonkey/issues/2675#issuecomment-349675249) for more background information.

### DIFF
--- a/src/bg/user-script-registry.js
+++ b/src/bg/user-script-registry.js
@@ -39,12 +39,7 @@ const db = (function() {
     });
   }
 
-  // Persist is a FF 57+ feature. Conditionally use it.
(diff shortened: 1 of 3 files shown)


Sxderp:
The failure has to do with the testing framework. `navigator.storage.persist()` never resolves when doing the tests. Therefore `openDb` is not called. Works in the browser / WebExtension context though.

Issue #2816, opened 2018-01-26
Native-like Monkey Menu needs fixes or roll-back


arantius:
The monkey menu after #2805 looks a lot better.  Except the little "next page" arrows look awkward when there are enough install scripts to trigger a scroll bar:

![screenshot from 2018-01-26 10 12 07b](https://user-images.githubusercontent.com/84729/35446149-d40f6008-0281-11e8-8ba6-341eeae58f89.png)

It seems like the scroll bar layers on top of, rather than beside, the content.

Sxderp:
RIP, I remember testing this and not having any problems. Must be platform / X environment dependent. Of course, aesthetic compatibility is ideal.

arantius:
This change needs either fixes or to be backed out.


See above problem, and issues mentioned in #2650, and I also get visual glitches on Windows.  On Windows every line seems much taller than on Linux, and if I scroll down I get:

![mm](https://user-images.githubusercontent.com/84729/35476309-ac800cf4-037b-11e8-8724-73e9ca15029f.png)

A bunch of non-erased garbage behind the real elements at the bottom.  If I scroll slowly, I can see each item, as it should disappear of the top instead pop in at the bottom, then disappear after a few further pixels of scrolling.


If I click into one script, the scroll bar keeps the height of the main view, giving me lots of nothing to scroll into.  If I do and scroll back up, I get black flashes.  If I do so slowly, they don't flash, I can get:

![mezer_01-27_16-04-28](https://user-images.githubusercontent.com/84729/35476320-e59569f8-037b-11e8-9388-6fd4d360e8ac.png)

Quite a lot of visual corruption.

Eselce:
Can confirm the scrollbar and the "garbage". Happened to me as well (win system)...

Sxderp:
The visual glitches aren't strictly caused by the changes. 

There's some Bugzilla posts about these issues. Of course, rolling back until Mozilla gets their shit together is always possible.

arantius:
Maybe I haven't looked closely before (testing profiles usually don't have enough scripts installed to scroll...) but even if the cause is a buggy platform, I can't recall seeing things like this before.

Sxderp:
> but even if the cause is a buggy platform, I can't recall seeing things like this before.

It's probably that the old styles didn't cause the issue to be prominent. I have a css fix that'll deal with the scrollbar, and somewhat resolve the graphical glitches, and inadvertently fix #2694 (lol).

[...]

Issue #2650, opened 2017-11-08
Firefox Nightly 58 - Overflow Menu width


Acid-Crash:
On latest Firefox Nightly 58 when button is placed inside Overflow Menu there is a graphical bug with menu width.
![gm](https://user-images.githubusercontent.com/32600318/32546511-0ee84e52-c488-11e7-93bd-c5d047dede4b.png)
It looks like menu isn't extending to full container width.

Sxderp:
The menu will somewhat be overhauled per #2639. I figure this will be fixed when those issues are addressed. I want to work on it but have recently been swamped.

Sxderp:
Turns out this was a very simple fix. Submitted a pull for it.

SW1FT:
This issue seems to be back in 4.3beta1.
![overflow](https://user-images.githubusercontent.com/6627739/35471994-82bee718-035e-11e8-8ca0-196321bb0398.png)

arantius:
> This issue seems to be back in 4.3beta1.

See #2816

arantius:
22fa2cda5a6149a9d695606a92302ed840964c21 closed this (but I typoed 2650 instead)

PR #2818, opened 2018-01-27
Monkey menu visual glitches

### LINKED ISSUE #2694: Maintain the popup scrolling position when viewing a script entry's options
Currently in GM 4.0 when clicking a script entry in the popup to view its options, and press the back '<' arrow,
then is always takes you to the top of the scripts list, regardless of your previous scrolling position.
I believe it would be more convenient if the scrolling position was maintained, 
especially when having lots of user scripts.

Sxderp:
Refixes #2650 
Fixes the scrolling issue of #2816 
Seems to reduce (doesn't eliminate) the ghosting effect in #2816
Inadvertently fixes #2694

### DIFF
--- a/src/browser/monkey-menu.css
+++ b/src/browser/monkey-menu.css
@@ -1,5 +1,6 @@
 :root {
   --arrowpanel-dimmed: hsla(0, 0%, 80%, 0.3);
+  --panel-max-height: 600px;
   --panel-separator-color: ThreeDShadow;
   --panelui-subview-transition-duration: 150ms;
 }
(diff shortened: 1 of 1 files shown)


Sxderp:
I should also note that the duplicated `(min|max)-width` attributes are necessary. If you just set `width` you get some seriously janky effects when navigating between the script details and installed scripts.

arantius:
> I should also note that the duplicated (min|max)-width attributes are necessary. If you just set width you get some seriously janky effects when navigating between the script details and installed scripts.

This is the sort of thing that should be called out via a comment right in line.  I noticed this myself as I tried to "fix" this, but could have easily missed its connection to the switching.

Sxderp:
> This is the sort of thing that should be called out via a comment right in line.

Added comments

arantius:
Looks good now, but #2822 to check cross platform.

PR #2820, opened 2018-01-28
Fix install dialog issues

### LINKED ISSUE #2815: Install dialog doesn't handle scripts with long names well
https://gist.github.com/arantius/3c513f82b7e5609f5ce3ae42b5bae53f

Install the no-space variant, see dialog something like:

screenshot from 2018-01-26 10 06 58

Where the name falls off the right side, causing other things like text in the rest of the dialog and (part of) the install button to also do so.

### LINKED ISSUE #2819: CSP preventing default icon from loading in install dialog
Not sure what we can do about this. When a dialog opens for a script that doesn't have a default icon you get the following error. UUID to be different for each person.
```
Content Security Policy: The page’s settings blocked the loading of a resource at self (“script-src moz-extension://17216af5-7fa6-43db-8011-285ae4f4de11”).
```

Sxderp:
Fixes #2819 
Fixes #2815

### DIFF
--- a/src/content/install-dialog.css
+++ b/src/content/install-dialog.css
@@ -9,11 +9,16 @@ html, body {
   overflow-y: auto;
 }
 html {
-  margin: 0 1em;
+  height: 100%;
 }
 body {
-  margin: 1em auto;
-  max-width: var(--dialog-max-width);
+  display: flex;
+  flex-direction: column;
+  justify-content: space-between;
+  height: 100%;
+  margin: auto;
+  padding: 1em 0;
+  width: var(--dialog-max-width);
 }
 
 body.result #install,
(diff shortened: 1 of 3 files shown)


arantius:
This breaks the layout of the dialog.

Before: [screenshot from 2018-01-30 09 49 31](https://user-images.githubusercontent.com/84729/35572584-ef95904e-05a2-11e8-8ca5-431d78a36c32.png)  
After: [screenshot from 2018-01-30 09 49 15](https://user-images.githubusercontent.com/84729/35572595-f46692e4-05a2-11e8-927a-38496728cdd5.png)

Sxderp:
> This breaks the layout of the dialog.

Fixed

Issue #2827, opened 2018-02-02
Test install in Firefox for Android


arantius:
I believe it's impossible to install scripts after #2719 in Firefox for Android.  Confirm.

Sxderp:
Reasoning? I haven't checked but I don't see anything that would imply it wouldn't work. According to MDN the new API calls are available in FF for Android.

I'll take a look and report back.

Sxderp:
The script dialog pops up fine and 'seems' to install.

I say 'seems' because there's a different problem. The Monkey Menu is borked. I'll hook up the debugger and see if I can figure it out.

Sxderp:
Actually, what breaks Android is #2812 

```TypeError: navigator.storage is undefined```

According to MDN: https://developer.mozilla.org/en-US/docs/Web/API/NavigatorStorage/storage
> Firefox 51 introduced NavigatorStorage, Navigator.storage, and the StorageManager.estimate() method. These are enabled by default in nightly builds but disabled by default in all other builds. You can enable these features by setting the preference dom.storageManager.enabled to true.

Looks like I'll have to re-add the conditional.

arantius:
Tested in 4.3beta4, works well.

PR #2828, opened 2018-02-03
Restore conditional persistence

Sxderp:
Android does not support persist

https://github.com/greasemonkey/greasemonkey/issues/2827#issuecomment-362747577


Eselce:
Quick style question:

In which cases should we use `async function scriptStoreDb(uuid)` (or `async function openDb()`).
It took me some unnecessary time to figure out, that `return openDb()` was okay (a returned `Promise`).

Just asking for a friend...

Issue #2823, opened 2018-01-31
Greasemonkey redownloads script for installation without valid cookies


mkeller-ergon:
We have some internal greasemonkey scripts hosted in an authenticated environment.
When clicking on a script, greasemonkey seems to get the correct contents and displays the URLs and other parameters correctly. However when selecting the Install-Button, it appears that Greasemonkey redownloads the file again, not in the context of the browser. This results in an installed user script that contains the HTML Login form instead of the expected contents.

This basically is even a huge security bug. If a hoster watches the source IPs closely, he could deliver the correct script upon the first request and deliver **ANY** other malicious script upon the second attempts which gets installed blindly by greasemonkey.

Instead, greasemonkey must use the same source it displays also for installation and not try to redownload the script itself again.

Sxderp:
Correct.

I've been working on this. I've got a branch that uses the initial FF request as the script source and never initiates a second request. It needs work and various other bits and bobs, I haven't had the time to comb through it.

Some of the items I need to go through and address include code style, error handling, locales, and see if I can add a progress bar (since @arantius wants one). Though, I'm not sure the benefit considering the flow that I've set up does everything in the background, as such the install window is safely closed after pressing install. Once the script is installed, or an error occurs, a notification pops up.

Since script resources are preemptively downloaded (before the user hits install the resources are fetched, aborting on cancel of course), I need to figure out what to do when the error occurs very soon in the download. That is, it occurs before the user can click install / cancel. When this happens they get a notification of error and then are left with a useless install window.

I kinda went on a tangent, but I am working on the issue.

Sxderp:
Might be relevant. https://bugzilla.mozilla.org/show_bug.cgi?id=1391992
The store information can then be passed to the install dialog and relevant cookies used in the downloader.

Issue #2817, opened 2018-01-26
Don't navigate to the script, when popping install dialog


arantius:
If I remember correctly, even 3.x acted this way.  We should still be able to, with WebExt APIs:

1. If navigation to any `.user.js` is detected, abort navigation.
2. Restart download of that URL in the background.
3. If that download results in a non-matching MIME, re-start the navigation -- filtered so we won't abort it.
4. Otherwise, pop install dialog and continue all downloads.

Sxderp:
I really think this should be rethought. Both VM and TM behave contrary to this. When navigating to a script you're presented with a page that includes both the script content/source and an installation button. While you are technically 'redirected' you are still, in essence, 'navigating' to the user script. Perhaps this should be brought up on the `-users` and `-dev` mailing list and see opinions others have?

arantius:
I support community discussion. 

Personally I'm of the opinion that ~nobody reviews the source, so it's a silly thing to show it to all users, including the majority who can't even understand it.

arantius:
And: reviewing only the main source, when there's also a `@require`, accomplishes very little.

StigNygaard:
I would be really sad if the possibility to review sourcecode doesn't return. I always feel more comfortable by skimming code. And if there are @require includes which doesn't point to a well known library (officially hosted jQuery etc), I'm always skeptical and usually aborts install unless I really need that script (in which case I skim the content of @require too).

But even if I am an exception and most people doesn't look at or understand the code, I think there's a preventive effect for script writers to know source is shown when installing.

In case you still decide not to show sourcecode, then please add easy accessible view-source: links to sourcecode (and preferable also any @require scripts included).

arantius:
As mentioned I also value review-ability.  For the minority of people who want it.  I just want it to work like #2567 so that you can review _all_ of the source.  You just click edit after install, and get all source and text resources visible.  Enable or uninstall as you choose.

Sxderp:
There's some performance implications on requiring multiple network requests. Specifically, between your `3.` and `4.` there's the "download to the point in which we have a valid metablock, so we can pop the install dialog".

Let's say the file isn't an actual user script and doesn't have a valid metablock. The installation dialog will never pop. The only course of action I see is to resume the navigation. Once navigation is resumed the file has to be redownloaded. An unnecessary request (#2830 addresses this by expanding what's currently being done in `onHeadersReceived`).

[...]

PR #2830, opened 2018-02-05
Move async script install and update to dedicated objects

Sxderp:
I've alluded to this branch for a while now (Ex. #2823). I feel like I've finally got it to a point where I can put it up as a PR. I expect there to be questions and concerns.

The two main things that this refactor accomplishes is:
1. Using the original Firefox request for script content (no second XHR for body). #2823 
2. Preemptive downloading of resources.

It took some work but I got those things to work with the progress bar on the install dialog. Although, once you press 'Install' the box can be safely closed without incident. The install / save will continue. A notification will alert you of failure or success. If you keep the window open then the errors (if there are any) will be displayed.

I don't expect this to make it for 4.3 or even 4.4, but hopefully somewhere around 4.5 some of this can be added in.


arantius:
> Although, once you press 'Install' the box can be safely closed without incident. The install / save will continue. A notification will alert you of failure or success.

I'm very anti "notifications".  On anything but the smallest screens, they often end up in a position that's easy to overlook.  More importantly, they usually (platform dependent) can appear and disappear with no interaction, leaving users in the dark.

I'd prefer that either the install button works and closes the window (success) or the window remains open and shows you, in context, what went wrong.

Sxderp:
> I'd prefer that either the install button works and closes the window (success) or the window remains open and shows you, in context, what went wrong.

If the dialog is open then it will report any errors. Although it doesn't auto-close on success. I wasn't sure if it should be left open and let the user close it themselves, or auto-close it.

I like the idea of 'fire-and-forget' in which you can click install and then close the window without issue. If you want to keep it open then all the information is still available to you. Along those lines, one improvement to the current code would be to not show notifications if the dialog window is still opened. That way you're not getting information twice (once in the dialog and another as a notification).

---

Rebased on master. Also, apparently I had missed a small, but important, commit[1] when I posted the initial PR. It has been added.

[1] "Calculate evalContent before saving a script"

arantius:
I have a strong preference for the entire install action to happen completely in one place: the one install dialog.  No other notifications, no other browser tabs, etc.  If it succeeds it closes (you're done with it).  If it fails, it's still open and shows you what went wrong.

Sxderp:
What's the intended behavior if someone hits 'install' but then closes the window (as far as I'm aware we cannot prevent this)? To me, that indicates an intent to install the script and a 'I don't want to sit around staring at a progress bar.' Should it be canceled? (RIP having to implement an abort option after the save intent has already been sent). That doesn't seem 'right' to me.

Eselce:
When I close an installation window, before I get a confirmation message, I surely don't want to finish that installation (unless I've used some `in background` button or else). That's rather, how I'd abort a zombie installation! Just saying...

Sxderp:
@Eselce , in GM 3.x you're forced to sit there staring at downloading progress bar before you can press 'install' [1]. And once you hit install you cannot 'backout'. You're committed (the window just closes). With what I'm suggesting that functionality remains, once you hit 'install' you're committed. However, I'm proposing the ability to hit 'install' before the download has completed. That way the window can be safely closed (if you choose) and you don't have to wait around.


[1] Use [this](https://gist.githubusercontent.com/arantius/3b493cb7ae6aad92d6c93881c14f1c4f/raw/83633b7630d66608f53290b4b6e2af813fc709c9/serve-user-script-slow.py) script. Curtious of @arantius . With GM 3.x navigate to `localhost:8000/script.user.js` and see that you're forced to wait.

arantius:
> GM 3.x you're forced to sit there staring at downloading progress bar before you can press 'install'

Because until the downloads are complete, they might fail.  As a user I certainly don't want to press install and then who knows when later discover that it didn't actually work. 

This is also how full extensions work, you have to wait for the download to complete and only then the "add" button appears.

Sxderp:
> Because until the downloads are complete, they might fail.

This applies to everything. I was putting user script install more akin to standard file downloads. When you download a file you're not forced to keep the window open watching the progress. And unless I'm remembering incorrectly they don't even notify you of failure.

Sxderp:
- Updated for `74bc5dcd3e533384e6adaa1f9961e4fbdfcfa427`.
- Changed to force users to stare at progress bars.

src/bg/objects/user-script-downloader.js
```diff
@@ -0,0 +1,182 @@
...
+        break;
+    }
+  }
+
+  _wrapFunction(fn, ...args) {
+    return data => {
+      if (data !== undefined) {
+        args.unshift(data);
+      }
+      fn.apply(this, args);
```
Eselce: Shoudn't a real wrapper `return` that value here? `apply` itself does...

Sxderp: Probably. I didn't think about it because all the wrapped functions don't return anything. Honestly, I probably don't need it in the `ScriptDownloader` class as it's only used in one place.

Eselce: I guessed so! No need for this in `onWhatever` callbacks, but unnecessary surprising, if used for other purposes...

Sxderp:
- Rebased for master.
- Although not strictly required, added `return` to the function wrappers.

Sxderp:
@arantius, I noticed you were working on some install-dialog stuff. However, I also noticed that you're in the middle of adjusting script downloading. In an effort to reduce duplicate work, I'd like to point out that this PR should be generic enough to work with pretty much any "detection / initial install / update" mechanism.

Once a decent way to interrupt the navigation is found and you restart the request (through XHR or whatever), you should be able to pass the content (through a promise) into the objects in this PR.

Sxderp:
Please checkout https://github.com/sxderp/greasemonkey/tree/async-script-install-update-2 for POC on the install dialog issue.

Sxderp:
Rebased for master.

Sxderp:
Fixed an issue with modifying and saving `@require` content.

Sxderp:
If this can get reviewed and pushed it will make #2817 doable with only some minor changes. Furthermore it sets the stage for adding more complex update features. Such as autoupdate.

arantius:
It's quite unlikely that this gets merged.

Sxderp:
Is there any reason? I thought it was better to separate script installation and updating rather than trying to keep everything in the registry and user script object. I had fixed your initial UI concerns, and any further ones I don't think should really be blockers. The main goal was to make installing and updating more generic requiring only the content and details (or promises to). And therefore it would be easy to adapt for different methods. And therefore any user interaction concerns should be doable without too much modification.

If you don't like the approach, well then that's fine too.

arantius:
The short version is that this PR is huge (+1,177 −601).  My equivalent (though not 100% done) [branch](https://github.com/greasemonkey/greasemonkey/compare/master...arantius:dialog-only-%232817#diff) is +602 -613.  We're generally removing similar things but you're replacing it with ~twice as much as me.  (I've also got at least some install dialog fixes that I noticed along the way; ideally I'll split those out separately and this change will shrink further.)

Why is there a user-script-downloader AND a user-script-updater AND a remote-install-manager AND user-script-installer?  They're each several hundred lines, but they're all doing basically the same thing.  Lots of places for bugs to hide, lots of things to need to read to understand what's going on.

It took me a while to get my branch to a decent place, but I very much like that it is doing downloads in the content (install dialog or editor tab) scope.  One, it's closer to the UI that has to know about the status and two, it presents a lesser possibility of long lived memory "leaks".

And so on.  Thank you very much for the effort, but I'm sorry about this one not working out.

Sxderp:
The `remote-install-manager.js` is used to handle communication between the install dialog and the background. If you're doing the downloading in the UI then that would cut ~200 lines.

The other files are in the vein of not overloading objects. Rather than having one class 'download and install' I have one that downloads and another that installs. As for having an installer and an updater, I had thought about it and figured updating a script and installing a script were sufficiently different to warrant different objects, rather than trying to fill the code up with additional conditionals deciding "does this need to be overwritten" or "what kind of input do I have", etc etc. Combining them might lead to another ~120 lines removed. Though, probably harder to read.

I also have additional code for aborting the downloads, which you don't have. Likely because everything is done in a content script, therefore the downloads should auto abort on tab close.

And probably some things here and there where I tend to use more whitespace or be a little more verbose, as in trying to account for as many problems or situations that may come up. For example our "individual" download classes [yours 56 lines](https://github.com/greasemonkey/greasemonkey/compare/masater...arantius:dialog-only-%232817#diff-61942cb841465b1190edca922487264cR200), [mine 91 lines](https://github.com/greasemonkey/greasemonkey/pull/2830/files#diff-cf4bffdb0145ac29d60f390f791b271aR91).

I don't know, it just seems like a mess to duplicate work. That's why I look at the network graph every now and again, to ensure I don't start working on something someone else is.

Anyway, if you don't want to use the code, okay.

PR #2835, opened 2018-02-07
Include cookies in GM.xhr for same-origin; implement withCredentials for cross-origin

### LINKED ISSUE #2826: Does GM.xmlHttpRequest send cookies?
ref:: FF59.0a1, Win65 GM4.3beta2

In a script that requires cookies (to stay logged in), the results of `GM.xmlHttpRequest` shows as a not-logged-in-user.

It seems that the cookies (session ones in this case) are not passed.
Is that the case?
How can I test it?
How can I fix it?

NoneGiven:
Normally, a background script XHR will only include cookies if third-party cookies are enabled in Firefox. These changes fix that using the `cookies` WebExtension API, giving `GM.xmlHttpRequest` cookie behavior identical to that of a normal XHR. Same-origin requests have cookies included automatically, and cookies can be included for cross-origin requests with the opt-in `withCredentials` parameter.

These changes only take effect if the user has not specifically passed a `Cookier` header to `GM.xmlHttpRequest`. If they do so, those values are sent and no further action is taken. This has the added benefit of allowing the user to opt out of automatic cookies on same-origin requests by manually passing an empty `Cookie` header.

If we don't want the cross-origin/withCredentials stuff, that can be easily removed (just remove `d.withCredentials` from the condition on line 127 of `on-user-script-xhr.js`).

Fixes #2826, I guess.

### DIFF
--- a/src/bg/on-user-script-xhr.js
+++ b/src/bg/on-user-script-xhr.js
@@ -123,14 +123,14 @@ function open(xhr, d, port, tabUrl) {
   
   // if same-origin XHR, add cookies unless already specified by user
   chrome.cookies.getAll({url: d.url}, cookies => {
-    if (!hasCookieHeader) {
-      if (d.withCredentials || isSameOrigin(tabUrl, d.url)) {        
-          var cookieStrings = [];
-          for (var cookie of cookies) {
-            cookieStrings.push(cookie.name + '=' + cookie.value + ';');
-          }
-          xhr.setRequestHeader(gDummyHeaderPrefix + 'cookie', cookieStrings.join(' '));
-        }
+    if (cookies.length && !hasCookieHeader
+        && (d.withCredentials || isSameOrigin(tabUrl, d.url))
+    ) {
+      var cookieStrings = [];
+      for (var cookie of cookies) {
+        cookieStrings.push(cookie.name + '=' + cookie.value + ';');
+      }
+      xhr.setRequestHeader(gDummyHeaderPrefix + 'cookie', cookieStrings.join(' '));
     }
     
     // if/when we switch from "chrome" to "browser" APIs, set up the promise
--- a/src/bg/on-user-script-xhr.js
+++ b/src/bg/on-user-script-xhr.js
@@ -105,11 +105,11 @@ function open(xhr, d, port, tabUrl) {
   d.responseType && (xhr.responseType = d.responseType);
   d.timeout && (xhr.timeout = d.timeout);
 
-  var hasCookieHeader = false;
+  let hasCookieHeader = false;
   if (d.headers) {
-    for (var prop in d.headers) {
+    for (let prop in d.headers) {
       if (Object.prototype.hasOwnProperty.call(d.headers, prop)) {
-        var propLower = prop.toLowerCase();
+        let propLower = prop.toLowerCase();
         hasCookieHeader = (propLower === 'cookie');
         if (gHeadersToReplace.includes(propLower)) {
           xhr.setRequestHeader(gDummyHeaderPrefix + propLower, d.headers[prop]);
@@ -121,15 +121,14 @@ function open(xhr, d, port, tabUrl) {
     }
   }
   
-  // if same-origin XHR, add cookies unless already specified by user.
-  // if/when we switch from "chrome" to "browser" APIs, set up the promise
-  // so we don't have to query cookies when we know we don't need them.
+  // If this is a same-origin XHR or the user opted in with withCredentials,
+  // add cookies unless already specified by the user.
   chrome.cookies.getAll({url: d.url}, cookies => {
     if (cookies.length && !hasCookieHeader
         && (d.withCredentials || isSameOrigin(tabUrl, d.url))
     ) {
-      var cookieStrings = [];
-      for (var cookie of cookies) {
+      let cookieStrings = [];
+      for (let cookie of cookies) {
         cookieStrings.push(cookie.name + '=' + cookie.value + ';');
       }
       xhr.setRequestHeader(gDummyHeaderPrefix + 'cookie', cookieStrings.join(' '));
@@ -151,12 +150,11 @@ function open(xhr, d, port, tabUrl) {
 
 
 function isSameOrigin(first, second) {
-  var firstUrl, secondUrl;
+  let firstUrl, secondUrl;
   try {
     firstUrl = new URL(first);
     secondUrl = new URL(second);
-  }
-  catch {
+  } catch {
     return false;
   }
   return firstUrl.origin === secondUrl.origin;
--- a/src/bg/on-user-script-xhr.js
+++ b/src/bg/on-user-script-xhr.js
@@ -15,11 +15,11 @@ function onUserScriptXhr(port) {
   if (port.name != 'UserScriptXhr') return;
 
   let xhr = new XMLHttpRequest();
-  port.onMessage.addListener(msg => {
+  port.onMessage.addListener((msg, src) => {
     checkApiCallAllowed('GM.xmlHttpRequest', msg.uuid);
     switch (msg.name) {
       case 'open':
-        open(xhr, msg.details, port);
+        open(xhr, msg.details, port, src.sender.tab.url);
         break;
       default:
         console.warn('UserScriptXhr port un-handled message name:', msg.name);
@@ -29,7 +29,7 @@ function onUserScriptXhr(port) {
 chrome.runtime.onConnect.addListener(onUserScriptXhr);
 
 
-function open(xhr, d, port) {
+function open(xhr, d, port, tabUrl) {
   function xhrEventHandler(src, event) {
     var responseState = {
       context: d.context || null,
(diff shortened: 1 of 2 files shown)


NoneGiven:
Note the unfortunate need to always query the cookies even if we know we won't use them. I find this would be easier to fix with promises (where we could use a dummy promise to continue without waiting for the cookies, or use the actual cookie promise if they are needed).

We would go from this:

```js
chrome.cookies.getAll({url: d.url}, cookies => {
  if (cookies.length && !hasCookieHeader
      && (d.withCredentials || isSameOrigin(tabUrl, d.url))
  ) {
    var cookieStrings = [];
    for (var cookie of cookies) {
      cookieStrings.push(cookie.name + '=' + cookie.value + ';');
    }
    xhr.setRequestHeader(gDummyHeaderPrefix + 'cookie', cookieStrings.join(' '));
  }

  var body = d.data || null;
  if (d.binary && (body !== null)) {
    var bodyLength = body.length;
    var bodyData = new Uint8Array(bodyLength);
    for (var i = 0; i < bodyLength; i++) {
      bodyData[i] = body.charCodeAt(i) & 0xff;
    }
    xhr.send(new Blob([bodyData]));
  } else {
    xhr.send(body);
  }
});
```

To this:

```js
var cookiePromise = new Promise(function(resolve, reject) { resolve([]) });
if (!hasCookieHeader && (d.withCredentials || isSameOrigin(tabUrl, d.url))) {
  cookiePromise = browser.cookies.getAll({url: d.url});
}

cookiePromise.then(cookies => {
  if (cookies.length) {
    var cookieStrings = [];
    for (var cookie of cookies) {
      cookieStrings.push(cookie.name + '=' + cookie.value + ';');
    }
    xhr.setRequestHeader(gDummyHeaderPrefix + 'cookie', cookieStrings.join(' '));
  }

  var body = d.data || null;
  if (d.binary && (body !== null)) {
    var bodyLength = body.length;
    var bodyData = new Uint8Array(bodyLength);
    for (var i = 0; i < bodyLength; i++) {
      bodyData[i] = body.charCodeAt(i) & 0xff;
    }
    xhr.send(new Blob([bodyData]));
  } else {
    xhr.send(body);
  }
});

```

I can't determine a way to do this with callbacks to the `chrome` version of the API that meets the goal of not actually querying the cookies if the conditions tell us we don't need them. If there is such a way, it's probably even messier than the current changes.

Eselce:
Just a detail: Is this

`var cookiePromise = new Promise(function(resolve, reject) { resolve([]) });`

not just another form of this?

`var cookiePromise = Promise.resolve([]);`

makyen:
There is no need to always query for the cookies, nor do you need Promises. You're having a problem because you haven't switched away from using an anonymous function once it became inappropriate to be using one. Just declare and name the function. Then, either do, or don't, do the query.

Something like:

    function sendTheXHR(cookies) { /*Code to send w/ or w/o cookies*/}

    if(noNeedToGetCookies) {
      sendTheXHR();
    } else {
      chrome.cookies.getAll({url: d.url}, sendTheXHR);
    }

& what @Eselce said re: `Promise.resolve([])`.

NoneGiven:
@Eselce Thanks!

@makyen I'm aware that's possible, but I wasn't sure about moving the send code to its own function. If @arantius is okay with it, then I'm okay with it, I suppose.

makyen:
@NoneGiven I'm confused. You said "I can't determine a way to do this with callbacks to the `chrome` version of the API that meets the goal of not actually querying the cookies if the conditions tell us we don't need them." Yet, then you said you do know how to do it, but that there is some unstated reason not to name a function. Creating a named function using just a `function` statement is effectively identical to creating a variable with `var`. You can also do so using a `function` expression, which can result in a named function by using `var`, `let`, or `const` to create a variable to hold the reference to the function. Thus, I'm unsure why there would be any problem with naming a function.

If it was that you just weren't thinking in that direction, I can understand not thinking of it. We *all* get focused on the way we're currently doing things, often just not thinking about some other way to do it, or write it. It often helps to come at the task fresh, either by getting fresh viewpoints from other people, or putting it aside for a while, so that you're looking at it with refreshed perspective. Having something like that happen is just part of being human.

Sxderp:
@NoneGiven, I haven't read the entire thread but if a anonymous function is getting large or causes code to become difficult to read then there's no problem separating it. And most people would probably not think much of it.

NoneGiven:
@makyen @Sxderp I feel like we're making a mountain out of a molehill. The point is just that I'm not a fan of moving the send code (which was already there, and not in an anonymous function) into its own function just because there's an asynchronous operation (getting the cookies) *which we don't always need* happening before it.

But I'm not strongly opposed to moving it out. Hence my saying I would defer to what @arantius likes better; it's his code.

NoneGiven:
I edited the original comment to more succinctly explain the issue with the flow around cookie querying. The best option for now seems to be to move the XHR sending code to its own method, which sucks but is probably necessary to avoid always querying for cookies.

If we were using promises instead of callbacks, we could accomplish he same goal without any additional functions/scopes by using await.

makyen:
@NoneGiven You appear to have a bias against callbacks and for Promises. While Promises can be quite beneficial, they aren't magic. You can have similar patterns with callbacks. Promises are syntactic sugar, which can be quite convenient and helpful, but they aren't the end-all-be-all of coding.

It's OK to have a preference for a particular coding style, or semantics, but you are allowing your bias to affect your thinking about options of how to solve the problem, particularly that you feel you *can't* do something without Promises.

The design flow that you've stated you want with Promises is: 

    var cookiePromise = Promise.resolve([]);
    if (!hasCookieHeader && (d.withCredentials || isSameOrigin(tabUrl, d.url))) {
      cookiePromise = browser.cookies.getAll({url: d.url});
    }

    cookiePromise.then(cookies => {/*Do all the things*/});

You could achieve exactly the same thing with callbacks doing something like:

    var cookieCaller = callback => callback([]);
    if (!hasCookieHeader && (d.withCredentials || isSameOrigin(tabUrl, d.url))) {
      cookieCaller = chrome.cookies.getAll.bind(null, {url: d.url});
    }

    cookieCaller(cookies => {/*Do all the things*/});

Given you are using `var` (i.e. not `let` or `const`) to define your variable, it's still hoisted to the top of your function-scope, just like if you had defined the main cookie function with `function`, assigning a name. As a result of this, there's really no scope difference between using a named function vs. defining a variable with `var` and using an anonymous function. If you were *really* concerned with scopes, then you'd use `let`.

The reason I didn't mention this design-pattern earlier, is that I consider this pattern to be undesirable in this instance. It's a bit more complicated than just naming the main cookie function. It goes out of it's way to avoid naming that function, sacrificing clarity and maintainability to retain that function as an anonymous function (with, or without using Promises). Just calling the function, or calling the function as a callback (or in a `then()`), is significantly cleaner.

If you were returning, or chaining things onto, the returned value from the call to  `cookiePromise.then()`, then the above pattern might be quite beneficial. But, that does not appear to be the case in this instance, as the result of that call to `then()` is not retained.

#### `await`?
Your comment regarding `await` isn't clear as to what you are saying the benefit would be, nor how you would implement it. While I can visualize a few different methods of using `await` in this instance, none appear to make the code clearer, or easier to follow, vs. just using Promises.

arantius:
> I edited the original comment

Please don't do that.  When comments are edited, it makes things unnecessarily difficult to understand; if I could disable this I would.  It's hard enough to read a serial log of comments and understand the reply chain; when messages change again later it becomes near impossible.

src/bg/on-user-script-xhr.js
```diff
@@ -118,18 +120,46 @@ function open(xhr, d, port) {
...
+  });
+}
+
+
+function isSameOrigin(first, second) {
+  var firstUrl, secondUrl;
+  try {
+    firstUrl = new URL(first);
+    secondUrl = new URL(second);
+  }
```
arantius: should not be a newline between the brace and the catch

src/bg/on-user-script-xhr.js
```diff
@@ -118,18 +120,46 @@ function open(xhr, d, port) {
       }
     }
   }
+  
+  // if same-origin XHR, add cookies unless already specified by user.
```
arantius: All comments should be grammatically correct sentences, including capitalization.

src/bg/on-user-script-xhr.js
```diff
@@ -105,10 +105,12 @@ function open(xhr, d, port) {
   d.responseType && (xhr.responseType = d.responseType);
   d.timeout && (xhr.timeout = d.timeout);
 
+  var hasCookieHeader = false;
```
arantius: Everywhere: `let`, not `var`.

NoneGiven:
@makyen The obvious reason Promises would be better in this case is because if anything should go into its own method, it's the cookie code, not the sending code, and we can't accomplish that with callbacks. The reason for using await would be to avoid extra nested scopes.

src/bg/on-user-script-xhr.js
```diff
@@ -105,11 +105,11 @@ function open(xhr, d, port, tabUrl) {
...
   d.timeout && (xhr.timeout = d.timeout);
 
-  var hasCookieHeader = false;
+  let hasCookieHeader = false;
   if (d.headers) {
-    for (var prop in d.headers) {
+    for (let prop in d.headers) {
       if (Object.prototype.hasOwnProperty.call(d.headers, prop)) {
-        var propLower = prop.toLowerCase();
+        let propLower = prop.toLowerCase();
```
Eselce: `const`? `propLower` has no other purpose than being the lowerCase version of `prop`!

revolter:
Could you please post an example? I don't see the `withCredentials` documented [here](https://wiki.greasespot.net/GM.xmlHttpRequest).

Issue #2837, opened 2018-02-09
Cannot install some scripts from openuserjs.org


chenjiajia9411:
Firefox Version:Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:60.0) Gecko/20100101 Firefox/60.0 ID:20180209102946
Greasemonkey Version:4.3beta4

When I tried to install [this script](https://openuserjs.org/scripts/joshu/BGM_TV_Bug_Fixer),the browser just returned a blank page and no installation popup turned up,also appeared when I attempted to install another [script](https://openuserjs.org/scripts/ruocaled/0.33333333333333333333333%E7%AD%89%E4%BA%8E1).

I also tried a brand new firefox with only greasemonkey installed but have no effect.

But if i disabled greasmonkey,the content of the script would show in the browser,so I guess this is a bug of greasemonkey.

Sxderp:
Well, it's a partial bug. The two scripts you link have invalid `@match` patterns. The intended behavior is to just show the script. However, you will not get an install dialog. And I don't think `@match` will be extended to accept the invalid patterns. I don't know if there's an official 'standard' for those patterns, but they're used basically is if there was one.

arantius:
> The intended behavior is to just show the script. However, you will not get an install dialog.

I disagree.  The script should be installed as if that `@match` wasn't even there.  Regardless of what kind of parse errors there might be, anything that isn't completely fatal should be ignored.

Sxderp:
Even if it's installed, you still get hit with the script not behaving as intended. In this particular case since the only match / include keywords are invalid it would be treated as a global match. And the script would run everywhere. I believe this is a worse than not installing.

PR #2838, opened 2018-02-10
Catch parse errors in detect-script

Sxderp:
See #2837

### DIFF
--- a/src/bg/user-script-detect.js
+++ b/src/bg/user-script-detect.js
@@ -37,7 +37,7 @@ function checkHeaders(responseHeaders) {
 
 // Check if enough content is available to open an install message
 function checkScript(userScriptContent, url) {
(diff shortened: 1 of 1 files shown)


Eselce:
Funny, this is a second view of my comment, no longer connected to the code...

Clarification:
```
function catchParseUserScript(userScriptContent, url) {
  try {
    return parseUserScript(userScriptContent, url, true);
...
```
The third parameter is just `true`, and the new function is missing a third parameter,
although it is called with three parameters (like plain `parseUserScript` call before).
Seems weird to me...

Sxderp:
Just an oversight. The call to `catchParseUserScript` should not have the third argument.

Sxderp:
I think I'll also add a console log, that way the errors don't go completely silent.

PR #2840, opened 2018-02-11
Add missing _() function to API scope

Eselce:
Several GM.* API functions are using _(...) for localized error messages.
A reference to the function in `src/i18n.js` is missing.
So a private copy of that code had to be imported to the API header.


Eselce:
Solves https://github.com/greasemonkey/greasemonkey/pull/2770#issuecomment-364648206

Eselce:
Also solves same issue #2844

PR #2842, opened 2018-02-11
Use a localized name for a new unnamed script (with placeholder $1)

Eselce:
A new unnamed script uses `// @name Unnamed script`, followed by a number.
This change makes it possible to use a localized name instead ($1 is the random number).


Eselce:
Instead of just appending, the new (english) message string is now inserted in alphabetic order...

PR #2845, opened 2018-02-12
Remove duplicate loadUserScripts function

Sxderp:


arantius:
How embarrassing.

src/bg/user-script-registry.js
```diff
@@ -144,7 +120,7 @@ async function loadUserScripts() {
           await saveUserScript(userScript);
         }
       }));
-      resolve(); 
+      resolve();
```
Eselce: There's a lack of error handling here (starting at line 139):
```
await Promise.all(event.target.result.map(async details => {
```
The returned `Promise` is never used, so probably, there should be sth. like
```
try { ... } catch(error) { reject(error); }
```

PR #2846, opened 2018-02-13
extName/extDesc Localized name and short description of the project

Eselce:
Make base string parameters in the project's manifest translatable/localizable:
I.e. name (`__MSG_extName__`) and short description (`__MSG_extDesc__`) of the project.


Eselce:
These strings are kind of special, so they should probably remain at the top of the file...

Sxderp:
I would consider the name to be a "proper noun" (English term) and thus not be translated. The description is still fair game.

Eselce:
> I would consider the name to be a "proper noun" (English term) and thus not be translated.

I think, I know, what "proper noun" is (a kind of fixed value of a special meaning/name of something),
and it's not really an issue of translation, but localization. What is the localized package/project name
as a string or symbol(s). What is the key string/symbols in an AppStore? I've visited some GitHub
projects, and it seems quite usual to do it this way: One (localizable) value for the whole project...

> The description is still fair game.

Even more than the name. I would be fine with english, but when I see all the (in my case: german)
descriptions and documents around, we're quite used to have strictly localized user interfaces.
But there's only an english wiki/homepage, thus I didn't change any links. The long description
on the AddOn details page has to be considered as well (looks a bit weird, when mixed with english).

arantius:
It makes lots of sense to localize the description.  But the name?

Would you translate "Coca Cola"?  Or "Google"?

Eselce:
As I said before, it's not about "translation" but "localization". My german version is no different to the english one. It's a centralized set-once-used-multiple setting, that may (but will seldom) change. Maybe there's an asian style symbol, that represents "Greasemonkey" better than the latin word. I don't know. But there's a reason, that it's pretty usual, that the project's name is coded as a `__MSG_...name__` (sometimes for marketing reasons). And it doesn't hurt anyone, does it? You could say, that it's useless, if there's no real translation. But it prepares for all cases, we are not thinking of. Nonetheless, I've not tried to substitute all the headers with "Greasemonkey" in it (like the editor title) with this value, that would be too much for now!

arantius:
Some extensions' names are very utilitarian and/or include descriptions ("Turn Off the Lights for YouTube™ and Beyond", "User-Agent Switcher", "Bulk Media Downloader" -- just a random few from the top of AMO's list).  Some ("Greasemonkey") are not.

Eselce:
Maybe there's some "Greasemonkey Beta". Or am I just lost, thinking that this is preferable? (I guess, there's nothing like "Greasemonkey™"!)

Sxderp:
I disagree with @arantius. The biggest argument for localizing the name has already been said.

> Maybe there's an asian style symbol, that represents "Greasemonkey" better than the latin word.

Furthermore,

> Would you translate "Coca Cola"? Or "Google"?

http://assets.coca-colacompany.com/migration/www.coca-colaconversations.com/wp-content/themes/coke/images/2008/08/06/coke_olympic_delicious_happiness_ca.jpg

NoneGiven:
>Maybe there's an asian style symbol, that represents "Greasemonkey" better than the latin word.

Brand names (and similar proper nouns) are frequently written in Latin script inside of CJK texts. It would be completely standard for the "English" term `Greasemonkey` to appear in a CJK UI.

Various other languages with non-Latin scripts are like this, too. Translating the extension name is very strange.

Eselce:
> Translating the extension name is not the right move.

This is not a "translation", it's a localization (and a constant, that is set once, used often).

Ad languages with non-Latin scripts: These text (CJK) often contain latin transscriptions, that's correct.

That doesn't prevent them to have a full (own) set of arabic/hexadecimal numbers (`０１２３４５６７８９` / `ＡＢＣＤＥＦａｂｃｄｅｆ`), dot (`・`) and even full stop (`。`). They need fullsize characters to match all the other characters, that fit into a monosized square. Not counting non-arabic numbers like `〇一二三十` or `壱弐参四五六七八九拾百千万萬`, `Α΄Β΄Γ΄Δ΄Ϛ΄(ΣΤ΄)Ζ΄Η΄Θ΄` (*) or e.g. `א  ב  ג  ד  ה  ו  ז  ח  ט`. See `٠  ١  ٢  ٣  ٤  ٥  ٦  ٧  ٨  ٩`. Which reminds me of L2R vs. R2L.

There are hundreds of special reasons, you can't even think of. But the my main goal was: Make a flexible locale for those who need it, cause it doesn't harm anyone else. `Greasemonkey = Greasemonkey` will for sure be the used case.

(*) This reminds me of the fact, how the greek nowadays sometimes (not always) write proper nouns - have you ever seen a greek movie poster? I have seen several of these in Greece. It's quite funny to extract the well known actor names. Quick search: This one with `ΣΗΝ ΚΟΝΝΕΡΥ` (`Σην Κοννερυ`) as `ΤΖΑΙΗΜΣ ΜΠΟΝΤ` 007 (`Τζαιημς Μποντ` 007) is currently available on ebay:

NoneGiven:
Localizing the extension name, whatever the difference may be, is also not the right move.

Eselce:
> Localizing the extension name, whatever the difference may be, is also not the right move.

I got that, and I see, that I have a minority point of view here.

But then please explain to me, why so many other projects do it that way?
(I don't even know the answer myself!)

NoneGiven:
>why so many other projects do it that way?

All evidence I've seen is to the contrary.

https://github.com/adblockplus/adblockpluschrome/blob/master/_locales/zh_CN/messages.json

```
"name": {
    "message": "Adblock Plus"
},
```

```
"description": {
    "message": "Adblock Plus 已被超过 1 亿台设备使用，是世界上最受欢迎的广告拦截软件。"
},
```

Eselce:
> All evidence I've seen is to the contrary.

What's the point? I didn't say, that everyone is doing it that way.

Eselce:
Rebased (via transifex_workable_strings) to 4.3beta5, 4.3beta6 (resolved merge conflicts), transifex order

arantius:
Did this conflict with #2860?  Was about to merge, but it looks like too much is changing.

Eselce:
Is derived from #2860 by a merge!

Eselce:
> @Eselce added some commits 4 days ago
> ...
> Merge remote-tracking branch 'origin/transifex-workable-strings' into 91749ed

Eselce:
Try this:

https://github.com/greasemonkey/greasemonkey/compare/master...Eselce:localize-manifest-json

Eselce:
Rebased to version 4.3beta7

Eselce:
This is still / again the correct reference:

https://github.com/greasemonkey/greasemonkey/compare/master...Eselce:localize-manifest-json

`Files changed` above is the difference to a long gone version!

Issue #2831, opened 2018-02-05
'Script Save Error' is not localized


Sxderp:
[Here](https://github.com/greasemonkey/greasemonkey/blob/b681b991cf293c96aa0b8702faf0b8c9547d122c/src/bg/user-script-registry.js#L283)
Should be a simple fix.

PR #2850, opened 2018-02-14
Localized save button, id/version, save error, _() usage, german strings

Eselce:
- Save button title is localized just before editor window is created
- User script id and version are formatted localized
- Script Save Error was already mentioned in #2841
- At two locations, the usage of `getMessage` was replaced by `function _()`
- Some new and upcoming german strings were added

### DIFF
--- a/_locales/de/messages.json
+++ b/_locales/de/messages.json
@@ -147,9 +147,19 @@
         "message": "Fehler beim Speichern: Allgemeiner Fehler.",
         "description": ""
     },
-    "$1; Version $2": {
-        "message": "$1; Version $2",
-        "description": ""
+    "$ID$; Version $VERSION$": {
+        "message": "$ID$; Version $VERSION$",
+        "description": "",
+        "placeholders": {
+            "id": {
+                "content": "$1",
+                "example": "namespace/name"
+            },
+            "version": {
+                "content": "$2",
+                "example": "1"
+            }
+        }
     },
     "Save": {
         "message": "Speichern",
--- a/_locales/de/messages.json
+++ b/_locales/de/messages.json
@@ -1,14 +1,14 @@
 {
     "You should only install scripts from sources that you trust.": {
-        "message": "Sie sollten nur Skripte installieren, deren Quelle Sie trauen können.",
+        "message": "Installieren Sie dieses Benutzerskript nur, wenn Sie der Quelle vertrauen.",
         "description": ""
     },
     "Greasemonkey is active": {
-        "message": "Greasemonkey ist aktiv",
+        "message": "Greasemonkey ist aktiviert",
         "description": ""
     },
     "Matches:": {
-        "message": "Muster:",
+        "message": "Ausgeführt auf (@match):",
         "description": ""
     },
     "GM.openInTab: Could not understand the URL: $1": {
@@ -40,7 +40,7 @@
         "description": ""
     },
     "Runs on:": {
-        "message": "Wird aktiviert von:",
+        "message": "Ausgeführt auf:",
         "description": ""
     },
     "Uninstall": {
@@ -76,15 +76,15 @@
         "description": ""
     },
     "User scripts for this tab": {
-        "message": "Benutzerskripte für diese Seite",
+        "message": "Benutzerskripte für diesen Reiter",
         "description": ""
     },
     "GM.xmlHttpRequest: Received no details.": {
-        "message": "GM.xmlHttpRequest: Keine Abfrage-Details übergeben.",
+        "message": "GM.xmlHttpRequest: Keine Anfrage-Details übergeben.",
         "description": ""
     },
     "GM.notification: \"text\" must be a string": {
-        "message": "GM.notification: Parameter \"text\" muss ein String sein",
+        "message": "GM.notification: Parameter \"text\" muss ein String sein.",
         "description": ""
     },
     "Requires special APIs:": {
@@ -120,7 +120,7 @@
         "description": ""
     },
     "Malicious user scripts can violate your privacy, steal your data, and act on your behalf without your knowledge.": {
-        "message": "Bösartige Benutzerskripte könnten Ihre Privatssphäre verletzen, Daten stehlen oder ohne Ihr Wissen in Ihrem Namen handeln.",
+        "message": "Böswillige Benutzerskripte können private Informationen stehlen oder Ihren Computer übernehmen.",
         "description": ""
     },
     "$1 does not @grant method $2": {
@@ -128,7 +128,7 @@
         "description": ""
     },
     "Greasemonkey is disabled": {
-        "message": "Greasemonkey ist inaktiv",
+        "message": "Greasemonkey ist deaktiviert",
         "description": ""
     },
     "This is a Greasemonkey user script.  Click install to start using it.": {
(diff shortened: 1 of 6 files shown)


Eselce:
Solves #2831 as a "side effect".

_locales/de/messages.json
```diff
@@ -1,14 +1,14 @@
 {
     "You should only install scripts from sources that you trust.": {
-        "message": "Sie sollten nur Skripte installieren, deren Quelle Sie trauen können.",
+        "message": "Installieren Sie dieses Benutzerskript nur, wenn Sie der Quelle vertrauen.",
```
arantius: I highly suspect that your changes here will be overwritten, next time I pull data out of transifex.

_locales/en/messages.json
```diff
@@ -2,6 +2,7 @@
   "$1 - Greasemonkey User Script Editor": { "message": "$1 - Greasemonkey User Script Editor" },
   "$1 - Greasemonkey User Script": { "message": "$1 - Greasemonkey User Script" },
   "$1 does not @grant method $2": { "message": "$1 does not @grant method $2" },
+  "$1; Version $2": { "message": "$1; Version $2" },
```
arantius: What?

I think this needs some description, or named placeholders.

src/bg/user-script-registry.js
```diff
@@ -239,28 +239,26 @@ window.onUserScriptUninstall = onUserScriptUninstall;
...
   function onSaveError(error) {
     let message;
     if (error.name == 'ConstraintError') {
       // Most likely due to namespace / name conflict.
-      message = chrome.i18n.getMessage(
-          'User script save failed: script named $1 already exists in namespace $2.',
-          [JSON.stringify(userScript.name),
-           JSON.stringify(userScript.namespace)]);
+      message = _('User script save failed: script named $1 already exists in namespace $2.',
+                  JSON.stringify(userScript.name),
```
arantius: All continued statements should be indented by exactly four spaces.

src/content/edit-user-script.js
```diff
@@ -1,5 +1,7 @@
 // TODO: Search, replace.
-// TODO: Put name in title.
+
+// Change this title as soon as possible, but it won't change later.
+document.getElementById('save').setAttribute('title', _('Save'));
```
arantius: It's probably worth using rivets, like in the rest of our HTML?  I'm worried about one-offs like this piling up over time.

src/user-script-obj.js
```diff
@@ -173,9 +173,10 @@ window.RemoteUserScript = class RemoteUserScript {
   }
 
   toString() {
-    return '[Greasemonkey Script ' + this.id
-        + (this.version ? '; ' + this.version : '')
-        + ']';
+    const fullScriptId = (this.version
```
arantius: 1) This should be formatted like

    const fullScriptId =
        this.version ? _('$1; Version $2', this.id, this.version) : this.id; 

2) This is not an "id".  We already call very different things "id".
3) I might actually prefer:

    if (this.version) {
      return _('[Greasemonkey Script $1; version $2]', this.id, this.version);
    } else {
      return _('[Greasemonkey Script $1]', this.id);
    }

The localized strings are more self-evident.  And IMO more readable.

_locales/de/messages.json
Eselce: All these strings are fully in sync with the transifex data.

With one exception, I am still observing: `New user script ...` has a new key phrase without `SPACE` before `...`
And there are some additional phrases, that will be added to transifex, once the english original is uploaded properly. I'll have an eye on them! (sp?)

_locales/en/messages.json
Eselce: I'd prefer placeholders as well, does this work with transifex? Not even sure about the `description` part...

src/bg/user-script-registry.js
Eselce: Tried that, was ugly, but no problem, it's okay...

src/content/edit-user-script.js
Eselce: I don't know too much about rivets, but does that translate anything other than `textContent`? You know, when I have `<tag>{Save|i18n}</tag>`, that's not a problem, but with `<tag text="{Save|i18n}">...</tag>`, it should fail. And "__MSG_Save__" generally doesn't work in `html` files as well.

src/user-script-obj.js
Eselce: I like the third version best.

src/content/edit-user-script.js
Eselce: Another remark: This will probably move to the menu bar later, and when in `src/content/menu-bar/menu-bar.run.js`, it will be handled much easier:
```
addItem('execCommand', { text: _('Save'), cmd: 'save', key: 's' })
```

arantius: Oh, yeah I'm not sure that/how it will work in an attribute.  Okay!

Sxderp: Pretty sure `<tag text=" 'Save' | i18n ">...</tag>` will work.

Sxderp: Er, might need `rv-text` as the attribute name.

Eselce:
Requested review changes done.

src/content/edit-user-script.js
Eselce: > Pretty sure <tag text=" 'Save' | i18n ">...</tag> will work.

Tried this:
```
< div id="save" class="command-item" title=" 'Save' | i18n " >
```
That didn't work: it showed  just `'Save' | i18n`!

Sxderp: With `rv-` ? I remember reading this somewhere and was like "Oh that's cool". But I never noted down the exact syntax.

Eselce: Trying sth. with `rv-title=`...

Eselce:
4.4 ? I'd really like to see the reviewed german strings in 4.3...

src/content/edit-user-script.js
Eselce: @Sxderp Oops, overlooked your comment! So you've had the same idea. If I read the documents correctly, we should use a prefix `rv-` in front of the attribute `title` and specify the value `value | formatter`, which is obviously `'Save'|i18n` in this case. But it reacts, as if there was no `title="Save"` at all, it's just ignored...

Sxderp: Make sure you
- Add `rivets.bundled.min.js` (or whatever it's called) to the `.html`
- Add `rivetsformatters.js` (or whatever it's called) to the `.html`
- Add `i18n.js` to the `.html`
- Put `rivets.bind(document, {});` somewhere in the `.js`

Assuming I didn't miss anything it should work then. I did a very minimal test with some one-off HTML page.

Eselce: Great! I forgot the binding. But it works now!

arantius:
Ugh, git did a fast forward instead of a merge (so I can't link to the merge I want to), but already pushed, so please a new PR for any further enhancements past 9ddc2603ae9a42da7ca801915966938661d0967c.

Eselce:
Do I have to rebase or sth.?

Eselce:
I think, #2852 that's all, what has been missing...

PR #2852, opened 2018-02-16
Smarter solution for translation of attributes with rivets

Eselce:
This is PR #2850 continued! The last commit was missing during merge...


Eselce:
This was part of the already merged #2850.
- Is there still a problem with it? (like fast-forward)
- Is the bind call okay? Do I have to read some more documentation, although it works?
- The rest seems straight-forward to me...

PR #2855, opened 2018-02-17
Only use notification port if it exists.

### LINKED ISSUE #2848: Minor notification/port issue
During some of my tests, I got some console messages from e.g. `shoelace.css`, `font-awesome.min.css` and `codemirror.css`, but more interesting, there was an unsynced port/channel. Pretty much every notification window close produced:
```
port is undefined  on-user-script-notification.js:50
```
and only once, I got (same line):
```
Attempt to postMessage on disconnected port  on-user-script-notification.js:50
```
The source code was:
```
47: chrome.notifications.onClosed.addListener(id => {
48:   let port = portMap.get(id);
49:   portMap.delete(id);
50:   port.postMessage({type: 'ondone'});
51:   port.disconnect();
52: });
```
I didn't really analyze it yet, but here I use Win 8.1, FF 59.0b9 (64 Bit), GM 4.3beta4...

Sxderp:
fixes #2848

### DIFF
--- a/src/bg/on-user-script-notification.js
+++ b/src/bg/on-user-script-notification.js
@@ -40,15 +40,19 @@ chrome.runtime.onConnect.addListener(onUserScriptNotification);
 
 chrome.notifications.onClicked.addListener(id => {
   let port = portMap.get(id);
-  port.postMessage({type: 'onclick'});
+  if (port) {
+    port.postMessage({type: 'onclick'});
+  }
 });
 
 
 chrome.notifications.onClosed.addListener(id => {
   let port = portMap.get(id);
-  portMap.delete(id);
-  port.postMessage({type: 'ondone'});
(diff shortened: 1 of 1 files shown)


src/bg/on-user-script-notification.js
```diff
@@ -40,15 +40,19 @@ chrome.runtime.onConnect.addListener(onUserScriptNotification);
...
+  }
 });
 
 
 chrome.notifications.onClosed.addListener(id => {
   let port = portMap.get(id);
-  portMap.delete(id);
-  port.postMessage({type: 'ondone'});
-  port.disconnect();
+  if (port) {
```
arantius: Rather than indenting ~the entire method, I'd prefer `if (!port) return;` as a general pattern.

PR #2858, opened 2018-02-17
Display a modal when saving a user script in the editor

### LINKED ISSUE #2847: How to avoid 'Script save error'
I do not know if it happened to me alone, but currently it is impossible to work on the scripts since it always gives the error 'Script save error'

I only can save the script one time, the other times give me the error when save  and I need to close the browser and reopen it. Obviously this is a sh**...

Is there any way to prevent the error from occurring?

Windows 8.1 x64
FF 58.0.2 64bit
Greasemonkey 4.3beta4

Sxderp:
fixes #2847


arantius:
FYI it's very helpful for me when you put `Fixes #NNN` in the commit message, not (only) the PR description.  Then GitHub automatically links things at merge time.

src/bg/user-script-registry.js
```diff
@@ -290,6 +290,14 @@ async function saveUserScript(userScript) {
       onSaveError(e.target.error);
       reject(e);
       db.close();
+    } finally {
+      // Send the script change event, even though the save may have failed.
```
arantius: This worries me.  If the save failed, I don't want the editor changing, and maybe throwing away unsaved details.

Sxderp: The `userScript` that's being saved should represent the editor as it is. You're basically returning the same details you submitted to the background, along with any downloaded resources. You shouldn't lose any unsaved details as `userScript` represents all those unsaved details (that failed to save).

Sxderp: Hm, I suppose that if the external resources take a long time to download time you may be able to perform some edits before the async save is returned. But that issue applies to both successful saves and unsuccessful saves.

I think the best thing to do is popup some modal that says 'saving' and doesn't go away until the result is back. If a save error occurs then display the error in the modal. If no errors, then auto close.

Sxderp:
Updated with a different approach.

arantius:
I see in an email update (but can't find anywhere else) that @Sxderp said:

> I think the best thing to do is popup some modal that says 'saving' and doesn't go away until the result is back. If a save error occurs then display the error in the modal. If no errors, then auto close.

I agree.

arantius:
I plan on merging this, but after some other work of mine -- I'll have to resolve merge conflicts at that time.

While we're waiting: does CodeMirror not have something to do (possibly modal) dialogs already?  Maybe https://github.com/codemirror/CodeMirror/tree/master/addon/dialog ?

If not is there a tidy general JS "modal dialog" component somewhere that we could simply reuse, rather than writing support for it from scratch?  A few moments of searching finds me:

* https://github.com/tomloprod/tomloprodModal
* https://madprops.github.io/Msg/
* https://github.com/AlexisTM/modalise.js

Which seem decent.  The smaller/lighter the better IMO.  Anything pre-existing with a documented API is good.

Eselce:
That was here: https://github.com/greasemonkey/greasemonkey/pull/2858/files/c6a35ab736815129a7e54263b69899475edcb276#r168963476

Sxderp:
If you want simple, I don't think you could get simpler than the code I wrote. It just uses rivets for the content and displays / undisplays by changing the body class (like the Monkey Menu).

Is it necessary for a "full feature" modal of some sort?

Sxderp:
When I have time I'll work on this for the recent changes to master. Should be done by Monday.

Sxderp:
I'm going to wait on this until we have better error handling in the downloader.

Sxderp:
Closing this and reopening as a new branch / PR

PR #2859, opened 2018-02-17
Set script name on editor tab when saving

Sxderp:
Requires #2858


Sxderp:
Going to close this. I'll post a new one when the script editor works properly.

Issue #2853, opened 2018-02-16
Update Translations with transifex


Eselce:
I've updated one moved and 7 new german strings in transifex.com.
Thanks for uploading the english original (not too late).

Some remarks:
- git doesn't like last lines without newline, thus I've added `\n` at the end
- What about the placeholders? There are none in the commit, can transifex handle them?
- Thus I get the wrong output:
```
Benutzerskript nicht ausführbar: [Greasemonkey script ; version ]
Error: executeScript requires either a 'code' or a 'file' property, but not both
```
or (with the correct strings from transifex.com):
```
Benutzerskript nicht ausführbar: [Greasemonkey Benutzerskript ; Version ]
Error: executeScript requires either a 'code' or a 'file' property, but not both
```
The placeholders are substituted with empty strings.

When I add the placeholder part to the entry:
```
    "[Greasemonkey script $id$; version $version$]": {
        "message": "[Greasemonkey Benutzerskript $ID$; Version $VERSION$]",
        "description": "",
        "placeholders": {
            "id": {
                "content": "$1",
                "example": "Name des Benutzerskripts"
            },
            "version": {
                "content": "$2",
                "example": "1.0"
            }
        }
    },
```
it works just fine:
```
Benutzerskript nicht ausführbar: [Greasemonkey Benutzerskript null/Unbenanntes Skript 500975; Version 1]
Error: executeScript requires either a 'code' or a 'file' property, but not both
```

arantius:
I tried re-writing the pull to do better/easier translation and I got at least one detail wrong.  I think there's some upper/lower case place holder issues?

Eselce:
Doesn't matter, it works this was (the key was `$id$` and `$version$` while the text was `$ID$` and `$VERSION$` and the placeholders `id` and `version`. I've tested that, it works (but maybe the same case would be better).

Eselce:
The only placeholder issue right now is, that the downloaded file doesn't contain any placeholders, and even more not translations for them. The placeholders in the strings are substituted by empty strings!

Sxderp:
Perhaps do without named placeholders? I can understand for longer strings or have more substitutions but this one seems fairly simple.

Eselce:
These placeholders were requested by @arantius - and there are even some more examples left. Time to solve the general problem...

arantius:
[...]

PR #2860, opened 2018-02-18
Transifex workable strings

Eselce:
Current transifex downloads for german (`de`), japanese (`ja`) and chinese (`zh_CH`) strings.
Adapted to work fine as `GM` locales. That is, with placeholders and `CR` at `EOF`.

@esperecyan Please give me a translation for the example string `User Script Name Here`, so I can append it to this pull request.

@yfdyh000 There are 7 (8) new strings to translate in transifex.com


Eselce:
Handles #2853 for all active locales

arantius:
My plan/expectation is to (like always in the past) download translation data roughly at release build time.  I'm even thinking of no longer committing anything but the English values, doing translations *only* at build time.

I'm not interested in merging pull requests for every new string for every language at who knows how many times, between releases.  Do you have a strong reason to do all this work?

Eselce:
I'm pretty used to doing overview on translation stuff, and here are my thoughts:
- Translation/localization takes some time (especially, if you have a lot of helpers, who are not always online or ready).
- Thus it's important to upload the english original as fast as possible.
- You are right, that only the original file must be traced, the others should be just dependant.
- But are still some points in #2853 that are not yet solved for an automatic process.
- E.g. the placeholders stuff. They are not yet supported (and contain further strings, that my be translated: `example`). If they are not worked on after download, they won't work at runtime.
- As you didn't seem to be comfortable with this stuff, I created a branch to collect that stuff.

arantius:
Thanks for all the effort and notes!  I would like to support both non-English users, and the kind individuals submitting translations.  I feel like I used to know what I was doing and had a good workflow for old/XUL extensions, and it's just another thing I have to learn over from scratch in the WebExt world.

One issue that's always happened in the past is keeping the translations from some external tool in sync with the checked-in copy of them, and this seems to be happening even more, recently.  I'm just worried about accepting pulls to non-english strings here; I feel like I'll destroy that work next time I update the translations from Transifex (or maybe whatever tool we switch to -- do you have experience/suggestions here?).

Eselce:
Most of experience comes from overviewing the language stuff of a boardgame site (much more than 100 languages, tens of natural languages and regular support for at least 12 natural languages, including CJK: german, english (uk/us), dutch, french, japanese, chinese (trad./simpl.), korean, spanish, italian, danish, russian, esperanto) for 125+ board games (since 1998) and its framework. The translation is done with a selfwritten Java support, but the issues are the same (even a similar concept: Key is the german source code, output without translations is detected automatically and can be marked as translatable).

You're totally right with worrying, that it's hard to stay up-to-date. Translators will come and go, communication channels (and tools) are vital. I'm positively surprised to have the 3 foreign language files updated within 2 days after your upload. That's a good start.

Tools: These are far from perfect. I don't see the possibility to "just download and install" these files without
reviewing and/or diffing the results before using them productively. The easiest way to do this was,
in my opinion, such a branch, created between the update cycles. Timeline here:
- 4.3beta1: First locale strings
- 4.3beta2: German (first draft)
- 4.3: German, Japanese, Chinese

It's also the easiest way to track the modifications. And to handle special things like the placeholder problem. It's good to have a proper workflow and process here. And to keep communication up (I didn't even know how to reach the other translators, unless they read the communications part of this repository).

Transifex seems okay for me, but it's not perfect. Honestly I have no suggestion for a different tool, that handles it all. Maybe the file type CHROME is not what we need, but what is the correct one? Are they extensible? I have still many questions here. Until then, manual work is required...

I don't really like automatic mass translations, but the advantage was, you could provide a bulk of supported (chrome) languages. That's fine, but human translations from natives will always be much better. What was it like in "ancient Greasemonkey times"? I think, I've only seen the german version (or was it english?)

For now, I think, one can handle the updates fine with transifex and manual diffing, unless you don't just copy the downloaded files into the repository. We could also introduce some more placeholders, that would not be much more work to do. It's on you to decide, I can only offer to provide some help. As a non-english native, this is a selfish task! ;-)

arantius:
> For now, I think, one can handle the updates fine with transifex and manual diffing, unless you don't just copy the downloaded files into the repository. 

Of course I just copy them, I don't speak those languages so I pretty much never know when any given change is good or bad.  But (I think) I trust the tool to at least generate a valid file which won't break anything.  (Whereas if I ask people to translate only via GitHub PRs, I expect to end up with invalid JSON sooner or later.  And I don't know whether I'd even know if one language or another will break, before I build.)

Eselce:
It's totally useful to leave it to (human) translators and transifex, instead of having many (bug prone) PRs!

Last month, german was the exception, because it was the only non-english version provided.
There should not be more than one PR for all foreign languages together.
The english original will be modified with the problem related PRs.

That doesn't mean, that you don't have to moderate the output, while using the strings as black box.
For example, the placeholders part of the files could as well be the same for every language.
And for me, it's often possible to do a facile check on most languages.

In my case, as a european, I have passive skills in german (native), english (school), french (school), italian (supported by latin), danish (holidays), spanish (holidays), greek (holidays), dutch (work, hobby). I've discussed a lot with our japanese, chinese and korean translators, which led to some basic knowledge about kanji, hiragana, katakana, hangul, etc. (imagine, how to moderate an argument between a young japanese and an old japanese, who didn't know english very well in the beginning, but a tiny bit of german, about japanese translations, they couldn't agree on - that was fun!) Not including most german dialects and minimal knowledge on derivates like scots (holidays), even some scottish gaelic, czech (holidays), catalan (holidays). Apart from the anecdotal quality of this, I want to make clear, that it's possible to handle languages without speaking them fluently (though english as a lingua franca helps a lot).

arantius:
There's now merge conflicts here, ugh.  See also #2862.

I'm now thinking even more of removing all but English from source control, putting the generated-by-translation-tool other locales only in the XPI builds.  Please argue for or against.

Eselce:
Already working on merge conflicts!

Eselce:
> I'm now thinking even more of removing all but English from source control, putting the generated-by-translation-tool other locales only in the XPI builds. Please argue for or against.

I've said, only tracking english sources is a good idea, but I rethought it, and it's not a good idea.
Not only because transifex is not perfectly reliable. It's good enough for holding the translation "intelligence".
It a useful interface for translators (output) and the source (input). This database is the reference for everything.

But it's not very good for tracking history (and annotations). When I tried to find out, which version of the german strings were actually in which version of 4.3betaN (N = 1, 3, 4), that was hard. Without the traces here in GitHub, it would be near to impossible. Moreover, when the time span from entering a translation to using it in a (even beta) release feels like eternal (it's just days and week, but I hope, you know what I mean). It's like, "when/why did I do this/that change? I don't remember!"

The current standings in the new 4.3beta5 is much crappier than the available GitHub/transifex data a day before. That tells me, that the procedure has to improve. Short term changes messed it up, translations were killed in the database, tracking is difficult, transifex order is totally mixed up and resorted, so that diffing is much more work. What's in your mind now will even increase these effects. Even for me, it's less work to maintain a single branch with updates than looking at the mess after such a release. It a bit frustrating! (Don't feel offended, in my region we are pretty direct and tell it like it is, not everyone will understand it, if not used to this!)

I will continue a bit of resorting/repairing now... TBC

Eselce:
Rebased to 4.3beta5, 4.3beta6 (resolved merge conflicts), transifex order

arantius:
I worry about what will happen when I download from Transifex, but for now: OK!

Eselce:
> I worry about what will happen when I download from Transifex, but for now: OK!

It's basically the same data (plus additional fixes) in the same order.
You'll be able to handle that (see diffs)!

PR #2871, opened 2018-02-24
The downloader ID is never used

Sxderp:


arantius:
Thanks, left over from the first implementation which lived only in the background and relied on more message passing.

Issue #2868, opened 2018-02-23
Monkey menu text sometimes cut off at bottom


stonecrusher:
*Moved from other [issue](https://github.com/greasemonkey/greasemonkey/issues/2740#issuecomment-367848276)*
![screenshot 3](https://user-images.githubusercontent.com/1388389/36568358-f683cf12-1829-11e8-9cce-ebfe09f30b4f.png)
For the cut off text, as written above:
`Greasemonkey 4.3beta5`
`Firefox 60.0a1 (2018-02-22) (64-Bit)` (happens with all three default schemes)
`Windows 8.1 home`, 125% scaling (should be also pretty standard because with 100% the text is *really* small. Can't remember having changed that.)
German localization for Windows and Firefox

Does not happen with Windows 7 home x32 (english), basic and classic colour schemes, 125% scaling (I think that's standard for nowadays resolutions and called "medium"), same GM and FF versions as above, just x32.

If you need further information, just ask back, I don't know what else could be helpful. But everything should be pretty standard because those are virtual machines with barely any software on it. OS is up to date.

I can't inspect the menu by myself, I always just get to the outer container `browser.webextension-popup-browser`.
Note that it seems that only the last item in a line doesn't have proper vertical alignment. So the text of scripts in the main menu are ok, just the corresponding rightarrow is shifted down. Accordingly new script and help item texts are shifted.
/edit
another screenshot about the last note
![screenshot 3](https://user-images.githubusercontent.com/1388389/36574390-eaf8b18c-1845-11e8-97f3-ed3902ae905c.png)

Eselce:
#2833

arantius:
Happens: Windows 8.1; doesn't happen: Windows 7.  This is ~impossible for me to test.  I have Win7 but not Win8 anywhere.

Where I am now (on Linux):

![untitled](https://user-images.githubusercontent.com/84729/36599252-d39e221a-187c-11e8-8946-c3bc85d4aaf8.png)

The text is 16px tall and exactly the same size as/aligned vertically with the icon.  Your text seems 
like it might be bigger, and is vertically offset.

https://wiki.greasespot.net/Troubleshooting_(Users)

Does this happen in a fresh profile?

Eselce:
Example:

Correctly aligned (list of 3 test scripts):

```
<div rv-hide="userScripts.inactive | empty">
      <hr>
      <div class="subview-item heading" rv-hide="userScripts.active | empty" style="display: none;">
[...]

PR #2872, opened 2018-02-25
Remove -moz-box and float from Monkey Menu

Sxderp:
This helps with #2868.
I don't want to say 'fixes' because the text is still off. See https://github.com/greasemonkey/greasemonkey/issues/2868#issuecomment-368278261

### DIFF
--- a/src/browser/monkey-menu.css
+++ b/src/browser/monkey-menu.css
@@ -119,7 +119,6 @@ section {
 
 .subview-back::before {
   content: url(/skin/back.svg);
-  display: -moz-box;
 }
 
 .subview-title {
@@ -146,14 +145,14 @@ section {
 }
 
 .subview-item {
-  -moz-box-align: center;
   background-color: transparent;
   color: -moz-fieldText;
   cursor: default;
-  display: -moz-box;
+  display: flex;
+  flex-direction: row;
   max-height: 24px;
   min-height: 24px;
-  padding: 4px 12px 4px 7px;
+  padding: 4px 12px 4px 10px;
   width: 100%;
 }
 
@@ -177,16 +176,22 @@ section {
   text-decoration: none;
 }
 
+.subview-item .icon::before {
+  vertical-align: middle;
+}
+
 .subview-item .icon {
   font-size: 14px;
+  flex-shrink: 0;
   text-align: center;
-  width: 19px;
+  max-height: 16px;
+  width: 16px;
 }
 
 .subview-item .icon img {
-  object-fit: contain;
-  width: 19px;
-  height: 17px;
+  max-height: 16px;
+  width: 16px;
+  vertical-align: unset;
 }
 
 .subview-item.disabled .icon {
@@ -194,17 +199,15 @@ section {
 }
 
 .subview-item .text {
-  -moz-box-flex: 1;
-  display: -moz-box;
-  overflow-x: hidden;
-  padding-left: 3px;
+  flex-grow: 1;
+  margin-left: 6px;
   text-overflow: ellipsis;
+  white-space: nowrap;
 }
 
 .subview-item::after {
   margin-inline-start: 10px;
   content: attr(after);
-  float: right;
   color: GrayText;
 }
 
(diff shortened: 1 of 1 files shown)


src/browser/monkey-menu.css
```diff
@@ -177,34 +176,38 @@ section {
...
   font-size: 14px;
+  flex-shrink: 0;
   text-align: center;
-  width: 19px;
+  max-height: 16px;
+  width: 16px;
 }
 
 .subview-item .icon img {
-  object-fit: contain;
```
arantius: This is what makes the icons retain the correct aspect ratio.  Are they still, when this is removed?

Sxderp: > This is what makes the icons retain the correct aspect ratio. Are they still, when this is removed?

Yes. As shown by @Eselce . And here's another image from a system that's not inflicted with the offset disease.

scaling

Sxderp: Scratch that. No it doesn't. Going to readd it shortly. Need testing from @Eselce to see if it screws up the lines.

Eselce: Damned: 1:2

Eselce: Lol, 2:2...

Eselce: Args, 3:3...

Eselce: 3:3 Overtime for Gold... Tell me, how to create that menu shot with such a distraction! :smile:

arantius: It would be a good idea to add two test scripts: one with an icon that is simply very large (wide and tall), one that is very wide (but not very tall).  And check both those cases.

Eselce: Wide is a long script name, but what is tall? Capital letters?

Sxderp: I think he meant the aspect ratio of the icon itself.

Eselce: My bad. Didn't read it carefully enough: "one with an **icon**..."
Does anyone have a script with such two icons?

Eselce: This is a little arbitrary, but...

Eselce: The latter icon is essentially of square shape (aspect ratio 1.0), the former is wider than the height.
Don't know, if it make sense to use icons of big data size (should be the same with huge icons of same aspect ratio)...

Sxderp: @Eselce , assuming those icons are free for use, can you make a PR adding them to the test scripts folder? If data URIs work for the `@icon` (if they don't then that should be addressed) then they can probably be included as such rather than separate files.

Eselce: Hmm., these were just arbitrary images, that I could find, maybe others are more useful.
They can be find on GitHub:

https://eselce.github.io/OS2.scripts/img/tor.png
https://eselce.github.io/OS2.scripts/img/sch.png

Hmm, funny, I thought, the goal image was not of square shape...

Eselce: There are probably better images available. Maybe I should do that test again!

arantius:
Pushed (47484e0ebde628adea6dd209d5587b1ba85b1e17) test scripts, which work as expected before merge:

mezer_02-27_20-36-59

arantius:
But is not quite perfect after:

mezer_02-27_20-44-14

The wide icon is now baseline aligned.  That might be fine?  This isn't likely a common case.

Eselce:
Need another Win 8.1 screeny?

Sxderp:
> Need another Win 8.1 screeny?

Not at the moment. When I get some time to figure out the baseline issue I'll ping you.

Eselce:
Okay, installed both scripts, waiting for the sources...

Sxderp:
@Eselce , Was planning on going to bed, but figured I'd take a few moments to look at it. Turns out the resolution was something stupid. Anyway, should be good for a check now.

PR #2881, opened 2018-03-01
Remove async from promise constructor in downloader

Sxderp:
We should really move on switching to browser.

### DIFF
--- a/src/downloader.js
+++ b/src/downloader.js
@@ -84,10 +84,11 @@ class Downloader {
   }
 
   async install(disabled=false, openEditor=false) {
-    return new Promise(async (resolve, reject) => {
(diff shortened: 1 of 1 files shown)


Eselce:
The async executor is still an issue in AMO.

Sxderp:
> The async executor is still an issue in AMO.

I don't understand what you're trying to say or imply.

Eselce:
> >   The async executor is still an issue in AMO.
> 
> I don't understand what you're trying to say or imply.

I was going through the PRs and issues and tried to remark,
that these two issues should be added (merged) soon, because
the master version still has async executors given in a Promise
constructor. But they are not assigned any milestones.

PR #2882, opened 2018-03-01
Cleanup async functions in user script registry

Sxderp:
Flattens the code. Removes unnecessary async. Uses `.catch` rather than `try .. catch` (huh?). Removed all the duplicate `db.close()`, you only have to close the db after you finish specify the commands you want to execute. The db won't close until those commands complete.

Also, what's the point of "RemoteUserScript". It's effectively useless. Created ***JUST*** for the `id`? Seems silly to me to have a class that doesn't do anything.

### DIFF
--- a/src/bg/user-script-registry.js
+++ b/src/bg/user-script-registry.js
@@ -47,59 +47,65 @@ async function openDb() {
 ///////////////////////////////////////////////////////////////////////////////
 
 async function installFromDownloader(userScriptDetails, downloaderDetails) {
+  let remoteScript = new RemoteUserScript(userScriptDetails);
+
   let db = await openDb();
-  return new Promise(async (resolve, reject) => {
-    try {
-      let remoteScript = new RemoteUserScript(userScriptDetails);
-      let txn = db.transaction([scriptStoreName], "readonly");
-      let store = txn.objectStore(scriptStoreName);
-      let index = store.index('id');
-      let req = index.get(remoteScript.id);
-      txn.oncomplete = async event => {
-        let userScript = new EditableUserScript(req.result || {});
-        userScript.updateFromDownloaderDetails(
-            userScriptDetails, downloaderDetails);
-        await saveUserScript(userScript);
-        resolve(userScript.uuid);
-        db.close();
-      };
-      txn.onerror = event => {
-        console.error('Error looking up script!', event);
-        reject();
-        db.close();
-      };
-    } catch (e) {
-      console.error('at installFromDownloader(), db fail:', e);
-      reject();
-      db.close();
-    }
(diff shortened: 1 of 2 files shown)


Eselce:
The async executor is still an issue in AMO.

src/bg/user-script-registry.js
```diff
@@ -47,59 +47,65 @@ async function openDb() {
 ///////////////////////////////////////////////////////////////////////////////
 
 async function installFromDownloader(userScriptDetails, downloaderDetails) {
+  let remoteScript = new RemoteUserScript(userScriptDetails);
+
   let db = await openDb();
-  return new Promise(async (resolve, reject) => {
-    try {
```
arantius: What happens to these exceptions that are no longer being caught by anything?

Eselce: I don't quite understand:

`installFromDownloader` itself is async (returns rejected Promise, if error is thrown).
Inside the executor, `reject` is called, then `catch` and `throw` => see above. IIRC.

Sxderp: For this particular point there's a total of five statements that would ever get caught in the `try .. catch` block. Three of those statements take static arguments and will not throw [1]. As for the other two, if they do throw the error will be bubbled up the async / promise chain.

If the database operation fails then the returned `new Promise` rejects and the error is again bubbled up the async / promise chain. The same applies with any error that may occur in the `.then` blocks. When using async / promises, there's no need to catch errors if all you're going to do is reject the promise in the catch block [2]. Promise structures do that for you.

[1] This assumes that the static arguments are coded properly and this bars any uncontrollable encounters caused by bad Firefox code (unlikely we could do anything about that anyway).
[2] This only applies if you are not going to change the error that is rejected. If you wanted to reject with a different error than you'd have to catch.

Eselce: A bit of sample code:
```
(async () => {
  throw 'err1';
  return new Promise((resolve, reject) => {
    throw 'err2';
    resolve('OK');
  }).then(() => {
    throw 'err3';
  }).catch((err) => {
    throw 'err4: ' + err;
  });
})();
```
Results (3 cases: 1. whole code, 2. without `err1`, 3. without `err1` and `err2`):
```
/*
Promise (rejected: err1)
*/
/*
Promise (rejected: err4: err2)
*/
/*
Promise (rejected: err4: err3)
*/
```

Sxderp: @Eselce could you do ` ```javascript ` for the opening code tag, to get syntax highlighting?

Eselce: @Sxderp :whistle: always forget that... :grin:

Issue #2813, opened 2018-01-18
Show number of running scripts in badge


FranklinYu:
This is a feature request. Both Tampermonkey and Violentmonkey shows the number of running scripts in current page. I guess they did it by [`setBadgeText()`](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/browserAction/setBadgeText).

Sxderp:
While this is easily doable now, it becomes more difficult if we switch to #2663.

kkot:
It could be an option in preferences. I would also like to see if a user script is running on the current page or not.

Eselce:
TODO: Move updateBadgeByDetails(), when #2663 is implemented.
TODO: Feature should be optional and user configurable.

Issue #2663, opened 2017-11-13
Use programmatic content scripts?


arantius:
https://bugzilla.mozilla.org/show_bug.cgi?id=1332273

Mostly setting this up to make sure the bugzilla bug is tracked.  This could/should be useful for us, especially for `@document-start` support?  Look into this after the above bug is resolved.  If it's useful, gate on stable release of the feature and implement.

Sxderp:
Assuming I understand it correctly, could be useful in that GM doesn't have to figure out what scripts match on navigation, only when installing / modifying and then have Firefox do the rest.

Though, for the topic of `@document-start` does switching to [webRequest.onRequestStarted](https://github.com/greasemonkey/greasemonkey/pull/2635) have any side effects in regard to being more consistent? I don't know how that can be tested / checked.

arantius:
> ... for the topic of @document-start does switching to webRequest.onRequestStarted ...

Even if we can detect the execution time, we have to (async) beam the script content down, which arrives at some non-start time.  In theory if we register the script with Firefox, it will "just work"?

(I don't want to get in the business of pre-caching things in content scopes in the hopes that it will already be there at start time.)

kekkc:
https://bugzilla.mozilla.org/show_bug.cgi?id=1332273 was fixed in latest FF59 some hours ago:
> Keywords: dev-doc-needed
Summary: Pass initial options to content scripts without async messaging (contentScriptOptions equivalent) → Implement a contentScripts.register API

So we need an additional API to use this?

Sxderp:
> So we need an additional API to use this?

No. I don't think so. They where just modifying the description / summary in order to more clearly label what the bug was about. Documentation is needed though. But looking at the [test cases](https://hg.mozilla.org/integration/autoland/diff/54ce507fdd65/toolkit/components/extensions/test/xpcshell/test_ext_contentScripts_register.js) can give an idea on how this is implemented in a WebExtension.

As for Greasemonkey, I think this should allow for the removal of the `execute.js` file. And we just register the scripts when they're first loaded / saved.

Sxderp:
I took a quick look at this and it'll be a bit tricky in regard to matches and includes. Greasemonkey takes the 'OR' approach while Mozilla [content scripts](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/manifest.json/content_scripts#Matching_URL_patterns) take an 'AND' approach for matches + includes. This is doable but not very cleanly.

[...]

PR #2888, opened 2018-03-01
Show badge with number of running scripts

Eselce:
Refs #2813 (fixes it for current implementation as of 4.3beta7)

This can easily be adapted later, after switching to #2663...


Eselce:
There is still a weakness in changing the tab.

arantius:
Thanks for your efforts but FYI: As a user I do _not_ want this, so I won't merge until there's a settings/options/preferences place I can turn it off.

Also: this implementation looks broken?  It will display how many _ran_, on whichever tab last ran scripts.  If I switch to a tab where more/fewer ran, the number won't change.

Eselce:
Yes, that's the weakness, I was talking about. This wasn't even meant to be in a PR now, but references appeared in #2813. This needs some more work (and options). BTW: What exactly is `Pony`?

Sxderp:
> Also: this implementation looks broken? ...

Need to pass in `tabId` from the details object into the browseraction methods.

Eselce:
@Sxderp Already did that, needs a smoke test...

src/bg/execute.js
```diff
@@ -30,27 +30,48 @@ function executeUserscriptOnNavigation(details) {
...
 
-  let userScriptIterator = UserScriptRegistry.scriptsToRunAt(details.url);
-  let count = 0;
-  for (let userScript of userScriptIterator) count++;
-  if (count) {
+  const enabled = true;  // TODO: query ('EnabledQuery' not yet available).
+  let count = [[0, 0], [0, 0]];
+  if (enabled) {
+    for (let i = 0; i < 2; i++) {
+      for (let j = 0; j < 2; j++) {
```
Sxderp: Assuming I'm understanding this right then what you're calculating here is
```
(scripts active that run on url) / (total scripts that run on url)
```
and
```
(scripts active) / (total scripts)
```

The way it's written is awfully confusing with the multidimensional count array. It would probably be much better to create a new function within the registry that will tally everything and return an object with keys for each count. Since the registry has all the data it only needs to iterate once.

Sxderp: Actually, the registry yields the userscript object directly. If you want you can still do it outside the registry. Just ask for "all" scripts and then do the checks.
```javascript
for (scripts) {
  let runOn = script.runsAt(url);
  runActive += (runOn && script.enabled) ? 1 : 0;
  runTotal += runOn ? 1 : 0;
  totalActive += script.enabled ? 1 : 0;
  total += 1
}
```

Eselce: The registry is (for sure) a better place for counting this data. I don't like the counting (we don't neet the data, just the count), I don't really like that array, but I didn't like plenty of different similar filenames either. I wanted to avoid a change to the registry, but I should count there! This will not be immediately be merged, so there's time to do that.

The association is here:
```
        const url = (i ? details.url : null);
        const all = !j;
```
First parameter `i` is `detected[1]` vs. `all[0]` scripts, second parameter is `activated[1]` vs. `total[0]` scripts. `all` means `detected`+`ignored`, `total` means `activated`+`deactivated`.

Thus the first line contains the `activated` scripts: running vs. idle (but active) scripts.
The second line contains all the scripts: running vs. all (active or not) scripts.
It's always from the viewpoint "how many scripts are running out of the available set of scripts".

It would also make sense to transpose that order, but I liked this one.

------------------------------------------------------------------------

My example (before I nuked my database) was:

In the GM menu, I had 6 scripts `for this tab`, 2 of them were active, the other 4 were greyed.
I had 20 `other` scripts, 13 of them were active, the other 7 were greyed.

The output was
```
Greasemonkey
2 / 15 active
6 / 26 total
```
(you could alternatively give 2/6 and 15/26 as output)

Eselce: Your little `for` loop should also work fine, except that I would write
```
  const runOn = script.runsAt(url);
  if (runOn && script.enabled) runActive++;
  if (runOn) runTotal++;
  if (script.enabled) totalActive++;
  total++;
```

Eselce: Correcting your initial assumption:

> Assuming I'm understanding this right then what you're calculating here is
> 
> `(active scripts that run on url) / (all active scripts)`
> 
> and
> 
> `(total scripts that run on url) / (total scripts)`

Issue #2869, opened 2018-02-24
Dealing with errors in downloader


Sxderp:
Something is missing.

Before I get to the issues, I need to pose a question. Should all downloads be aborted if an error occurs? Some thoughts I came up with for both sides. Yes, they should be aborted, you're wasting resources downloading everything else when we already know the script is not going to be installed. No, they shouldn't be aborted, you can document additional errors that occur. Currently nothing is aborted.

The issues, if a download fails for some reason, then it will [throw an error](https://github.com/greasemonkey/greasemonkey/blob/7db5f084acc5be7cb79c5d8efc76fd35212f2bed/src/downloader.js#L157-L160) in `async start()` but there is nothing to [catch these errors](https://github.com/greasemonkey/greasemonkey/blob/7db5f084acc5be7cb79c5d8efc76fd35212f2bed/src/content/install-dialog.js#L113). Furthermore, it looks like there was some [tracker of errors](https://github.com/greasemonkey/greasemonkey/blob/7db5f084acc5be7cb79c5d8efc76fd35212f2bed/src/downloader.js#L10) (somewhat makes sense if downloads are not to be aborted) and it's even [checked for content](https://github.com/greasemonkey/greasemonkey/blob/7db5f084acc5be7cb79c5d8efc76fd35212f2bed/src/content/install-dialog.js#L17) [1], but nothing is ever added to this error list.

What it looks like is that the code can't decide if it wants to abort everything and report an error right away, or if it wants to sit around waiting for everything to possibly finish, tallying errors as they come in, and then report.  Currently, it does neither.

As an additional point, if `scriptDetails` is rejected then `async start()` will resolve without any errors. Which would imply that "all is well."

[1] This on progress error checking is broken in two ways. The first is that it loops over `message`? And the second is that, assuming `gDownloader.errors` ever had any content, if there was a single item still downloading and posting progress a new error list would be created and appended every time that progress was posted.

---

As an aside, it might be a good idea to split `async start()` up into multiple functions. It is a somewhat long.

Eselce:
Maybe this fits in here (not fully): If you change the @name of a script to an existing name and save, that error is no longer handled. And it's silent. I also used a new script, saved it, then there were two scripts resulting.

[...]

PR #2889, opened 2018-03-03
Differentiate Download Errors

Sxderp:
Continued effort to improve error handling in the downloader.
refs #2869 

---

The editor will also need improvements but that's for another PR.

### DIFF
--- a/src/downloader.js
+++ b/src/downloader.js
@@ -1,3 +1,13 @@
+
+class DownloadError extends Error {
+  constructor(failedDownloads) {
+    super('Download(s) failed; see `.failedDownloads`.');
+    this.name = this.constructor.name;
+    this.failedDownloads = failedDownloads;
+  }
+}
+
+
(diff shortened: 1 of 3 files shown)


Eselce:
i18n in DownloadError?

Sxderp:
I dunno. That particular string is more of an internal string. It's never displayed to the users anywhere. It's not even output to the console.

src/downloader.js
```diff
@@ -270,4 +268,14 @@ class ImmediateDownload {
...
 
+
+class DownloadError extends Error {
+  constructor(failedDownloads) {
+    super('Download(s) failed; see `.failedDownloads`.');
+    this.name = this.constructor.name;
+    this.failedDownloads = failedDownloads;
+  }
+}
+window.DownloadError = DownloadError;
```
arantius: If you want it to be public, just put it outside the scope that hides the private members.

PR #2899, opened 2018-03-08
Right click opens script in editor

Eselce:
This makes life easier...


src/browser/monkey-menu.js
```diff
@@ -158,6 +158,29 @@ async function onHashChange(event) {
 }
 
 
+function onClick(event) {
+  if (event.which === 3) {  // Right mouse click.
+    if (!gActiveUuid) {
+      const elements = document.querySelectorAll(':hover');
+      // Last element is the name of the script; last but one is the anchor.
+      const hash = elements[elements.length - 2].hash;
```
Sxderp: Is query selector deterministic? It might be safer to go up from `event.target` looking for `a` tags. Perhaps [`el.closest()`](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest). Or a loop like:
```javascript
while ((el = el.parentNode) && el.tagName !== 'A')
// Will have null or the element. Bail if null.
```

Sxderp: Slight problem with my while loop. Flip the parameters around and use `.parentElement` instead.
```javascript
while (el.tagName !== 'A' && (el = el.parentElement))
```

Eselce: I didn't know how to obtain the anchor, so I googled a bit and tested several solutions given. Then I found this solution (for not-so-old browsers). It returns an array (rather a stack) of all elements from `<html>` over `<body>` to `<a>` and `#text`. Almost exactly what I wanted. IIRC `event.target` (which would be a perfect solution) didn't work. I should test that again...

Eselce: Oops, strange. Now it works! This is much easier:
`event.target.parentElement`

Sxderp: > `event.target.parentElement`

I don't think this is guaranteed to be the correct element. If I remember correctly clicks are first sent to the lowest node in the tree, that you're hovering over (which is the target). If you click over the icon then your parent node is going to be `div.icon`. The safest, and more robust in case of changes to the HTML, would be to do some sort of parental search.

Eselce: I wanted the area with the script name as target, so its parent is always the anchor.

Eselce: Hmm, okay that works fine with the icons. But now I need a check against null...

Eselce:
Rebased to master with new monkey-menu.

Sxderp:
I'd suggest making this middle mouse instead of right mouse. To follow the "middle mouse opens link in new tab" theme.

Eselce:
Would be a 1 bit change (0->1), and I thought about that, but
- I don't really like middle-button-activation, had often issues with lacking compatibility or usability
- Tampermonkey uses right-button, so I am quite used to it :smile:

ddgls08 (approved):

Cerberus-tm:
This would be really nice to have. I still miss right-clicking after all this time. It's how it always to work, and how I believe other userscript extensions do it, too. But is there a down-side to it?

roger21:
I am currently editing a bunch of scripts, not having that feature is quite annoying to say the least.

> it's non-discoverable

who care, it's not like the GM4 UI is top of the art nor accessible, i'm not saying it's bad, i'm saying is quite minimalistic and quite far from any full page UI with a full featured list of script with all the buttons directly available, so adding this would be a greatly appreciated small concession

>  It's impossible on some platforms.

that has to be the weakest argument either in that context, gm has only one target browser, and if your suggesting tablet device, then you have to be kidding : who would do editing work on a tablet and even though the left click is still available

> It's a very small enhancement

yes the kind of little bit of change that changes everything

PR #2900, opened 2018-03-10
Use save modal with editor

Sxderp:
see #2858

### DIFF
--- a/src/content/edit-user-script.css
+++ b/src/content/edit-user-script.css
@@ -47,14 +47,13 @@ body #modal {
   font-size: 25px;
   margin: 0;
 }
-#modal button {
+#modal header {
+  border-bottom: 1px solid #CCC;
+}
+#modal footer button {
   float: right;
   font-size: 15px;
   width: 140px;
-}
-
-#modal .panel-head {
-  border-bottom: 1px solid #CCC;
 }
 #modal .panel-body {
   font-size: 14px;
--- a/src/content/edit-user-script.css
+++ b/src/content/edit-user-script.css
@@ -7,11 +7,12 @@ html, body {
   font-size: 12px;
 }
 
-
-body.save #modal
-{ display: grid; }
-body #modal
-{ display: none }
+body.save #modal {
+  display: grid;
+}
+body #modal {
+  display: none
+}
 
 #modal {
   background-color: #A3A3A363;
--- a/src/content/edit-user-script.css
+++ b/src/content/edit-user-script.css
@@ -116,6 +116,7 @@ body, #editor {
 .CodeMirror {
   height: 100% !important;
 }
+/* Required due to inclusiong of shoelace.css */
 .CodeMirror pre {
   line-height: 14px !important;
 }
--- a/src/content/edit-user-script.css
+++ b/src/content/edit-user-script.css
@@ -1,9 +1,64 @@
 html, body {
   background: lightgrey;
-  height: calc(100% - 8px);
+  height: 100%;
   margin: 0;
   overflow: hidden;
   padding: 4px;
+  font-size: 12px;
+}
+
+
+body.save #modal
+{ display: grid; }
+body #modal
+{ display: none }
+
+#modal {
+  background-color: #A3A3A363;
+  grid-template-rows: 1fr 1fr 1fr 1fr;
+  grid-template-columns: 1fr 1fr 1fr 1fr;
+  position: absolute;
+  top: 0;
+  left: 0;
+  height: 100vh;
+  width: 100vw;
+  z-index: 100;
+}
+#modal .panel {
+  background-color: #FCFCF9;
+  border: 1px solid #CCC;
+  border-radius: 1.7px;
+  display: flex;
+  flex-direction: column;
+  grid-column: 2 / 4;
+  grid-row: 2;
+  padding: 1rem;
+}
+
+#modal .fa {
+  font-size: 25px;
+  margin-right: 10px;
+  vertical-align: unset;
+}
+#modal h2 {
+  display: inline-block;
+  font-weight: 300;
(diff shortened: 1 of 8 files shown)


arantius:
With this, every single save action causes the entire window to blink/flash, even if it completes "instantly".  I don't like that.  This is a strange behavior for a text editor.

(If I haven't added remotes, it should "just save" with no dialog.  If I have, then and only then it makes sense to show the progress of those downloads.)

Sxderp:
While I can see your point and I have some ideas on how to implement those checks (may require reworking of `Downloader` depending on how 'smart' the checks are). There's still the problem that we have no idea how long a "save" will take. I don't know how normal text editors deal with this problem (obviously not all of them save synchronously) but I can't think of any sane way (caching entire contents being the only idea I have) to deal with little things like marking the editor as "clean." You make some changes, save, saving takes a while and during that time you make more changes, the save completes, how do we figure out whether to mark the editor as "clean."?

I suppose we could mark it clean before the save and mark it dirty if there's an error. But that might also seem like bad UX.

arantius:
A "normal" editor I think always assumes that a filesystem write is always fast.  It doesn't have to plan to remotely download things over the internet as part of a save action.  I can't think of another text editor that ever has any kind of progress/dialog as part of a save action, that I've used.  (Also, neither VM nor TM do this.)

I think the simplest (best?) solution would be to delay showing the modal for some amount (100ms?) of time.  Enough that we expect that a plain save (no new remotes to download) will always have passed.  If that has passed and we're still saving, then show the dialog with its progress bar.  If not, just never show the dialog.  Either way, mark clean only when save has finished.

It's not too complicated on top of that to track clean/dirty of parallel save and edit actions.  Just a few cases to design in.

Sxderp:
> I can't think of another text editor that ever has any kind of progress/dialog as part of a save action

I've seen editors that save synchronously. Most of the time the editor just freezes rather than bring up a dialog box.

Sxderp:
Updated with a simple timeout before showing the save modal. Much easier than anything I had originally thought of (which involved doing checks to see if we actually needed to fetch remotes, etc.).

src/content/edit-user-script.css
```diff
@@ -1,9 +1,64 @@
...
+  height: 100%;
   margin: 0;
   overflow: hidden;
   padding: 4px;
+  font-size: 12px;
+}
+
+
+body.save #modal
+{ display: grid; }
```
arantius: Rarely all on one line, otherwise always:

    selector {
      rules
    }

src/content/edit-user-script.css
```diff
@@ -63,3 +118,6 @@ body, #editor {
 .CodeMirror {
   height: 100% !important;
 }
+.CodeMirror pre {
+  line-height: 14px !important;
```
arantius: Why? Does this have anything to do with showing a modal dialog?  Some side effect of now including shoelace?

src/content/edit-user-script.html
```diff
@@ -39,6 +40,36 @@
...
+      <div rv-hide="modal.errorList | empty">
+        <p>{'script_save_error'|i18n}</p>
+        <ul>
+          <li rv-each-error="modal.errorList">
+            {error}
+          </li>
+        </ul>
+      </div>
+    </div>
+    <div class="panel-foot">
```
arantius: Just `<footer>`?

src/content/edit-user-script.html
```diff
@@ -39,6 +40,36 @@
 </head>
 
 <body>
+<div id="modal">
+  <div class="panel">
+    <div class="panel-head">
```
arantius: Just `<header>`?

src/content/edit-user-script.html
```diff
@@ -39,6 +40,36 @@
 </head>
 
 <body>
+<div id="modal">
+  <div class="panel">
```
arantius: Why do I need an extra wrapper div here?

src/content/edit-user-script.html
```diff
@@ -39,6 +40,36 @@
...
+      <h2>{modal.title}</h2>
+    </div>
+    <div class="panel-body">
+      <div rv-show="modal.errorList | empty">
+        <progress id="modal-progress">
+      </div>
+      <div rv-hide="modal.errorList | empty">
+        <p>{'script_save_error'|i18n}</p>
+        <ul>
+          <li rv-each-error="modal.errorList">
```
arantius: This fits so should probably be all on one line `<li>...</li>`

src/content/edit-user-script.js
```diff
@@ -173,3 +217,4 @@ window.addEventListener('beforeunload', event => {
     event.preventDefault();
   }
 });
+rivets.bind(document, gTplData);
```
arantius: Should be blank lines, but why did this need to move all the way to the bottom?

src/content/edit-user-script.css
Sxderp: Slide effect of shoelace.

src/content/edit-user-script.html
Sxderp: The outer div covers the entire page and consists of a semi transparent dark color. Effectively "dimming" the contents. The inner div is the focused "box". If you have suggestion on a better way to do this, shoot.

src/content/edit-user-script.js
Sxderp: Not really sure why I stuck it all the way at the bottom. Should only need to be below "tplData"

Sxderp:
> Status check on this PR please?

Should be good. Resolved all the issues in the review.

Sxderp:
Rebased on master

Sxderp:
Rebased on master

PR #2903, opened 2018-03-12
Bind rivets earlier

Sxderp:


Sxderp:
Note to self: Determine if `JSON.stringify` is still required.

arantius:
Merged by hand.

Sxderp:
> Merged by hand.

I noticed you removed the "catch" on the `scriptDetails` promise. This will cause the uncaught error message in the console. Granted, I'm pretty sure it includes the error object and thus fill basically the same role as the `console.warn`.

The reason for the message is that download errors are rejected twice. Once for the `scriptDetails` promise and another for the `start()` promise. I suppose the `scriptDetails` rejection could be removed.

Sxderp:
An alternative, while semantically not as nice, would be to remove the (public) `scriptDetails` promise and use a progress listener. Checking if the details are ready and if so acting on that. Removing the progress listener after (would need a method to remove listeners  from the `Downloader`). That would avoid the double rejection.

PR #2904, opened 2018-03-12
Failing to find meta is not fatal

Sxderp:


arantius:
I like the last (apply defaults) commit here, but the rest seems perfectly backwards.  At the very least, the behavior no longer matches the names of things.  (In my mind at least, returning a value successfully is not "fail"ing, even if that value is null.)

arantius:
Just 5e9d369d271202d3fd5d9b99b8495cce38418622 and 893a63dc2dc171917ce6d6dfc41da6c59f4e6e37 address this, and leave the (parameter) names and values logically consistent.

PR #2913, opened 2018-03-18
Scroll only userscript section in Monkey Menu

### LINKED ISSUE #2809: RFE: Improve the Monkey Menu when many scripts are installed
Note; I have 100+ scripts in the menu

It could be useful to:

- Move "New  user script..." to top
- Add the scrollbar to "Other scripts" keeping the menus static

Sxderp:
fixes #2809

### DIFF
--- a/src/browser/monkey-menu.css
+++ b/src/browser/monkey-menu.css
@@ -131,6 +131,21 @@ section {
   width: 100%;
 }
 
+.subview-body hr.hr-top {
+  margin-bottom: 0;
+}
+.subview-body hr.hr-bottom {
+  margin-top: 0;
+}
+.subview-body div[style*="none"] + hr.hr-bottom {
+  margin-top: 6px;
+}
+
+.script-scroll {
+  max-height: 380px;
+  overflow-y: auto;
+}
+
 .subview-body {
   max-height: 560px;
   padding: 6px 0;
--- a/src/browser/monkey-menu.html
+++ b/src/browser/monkey-menu.html
@@ -25,39 +25,45 @@
     </div>
     -->
 
-    <div rv-hide="userScripts.active | empty">
-      <hr>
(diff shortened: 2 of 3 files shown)


arantius:
I'm not convinced I want this; got screenshots before and after?

Also: you can probably get away with much less code (and less changed) by switching the divs from `rv-hide` to `rv-if`, then using CSS to hide consecutive `hr`s?

Sxderp:
> Also: you can probably get away with much less code... 

I need to test the following, but I don't think so. Unless I'm mistaken rv-if is evaluated once. So the DOM won't update as you add or remove scripts. Thus you'd have a 

``` 
<hr> 
<div></div> 
<hr> 
``` 

Thus the `<hr>` are not consecutive.

Sxderp:
For note, right now the code scrolls all the scripts under one scrollbar. I just thought it might be decent to scroll the two sections individually. That way, say you have 100 scripts, you can know right away if the script ran on the page rather than having to scroll up to look at the heading.

Sxderp:
To clarify my above. "right now this PR..."

Eselce:
Appreciating the idea of scrolling only the user scripts (or even only the "other" user scripts)...

Sxderp:
Either both should be scrolled or none. For example if you have a lot of active scripts it seems weird to scroll the entire content to go through them but not have  to scroll the entire content for the other scripts (nested scrollbars suck).

Eselce:
All of which are active on the current page/tab? Okay, may happen, if you have many universal scripts (`@include *`).  For me, I normally have a max of 5 active scripts on a page, much more are inactive on that page (but activated).
As I said, all user scripts (both sections) being scrolled is a good option.

Sxderp:
Pictures

before-after

Eselce:
Okay, that's focused and clear enough...

PR #2916, opened 2018-03-25
Localize menu userscript details

Eselce:
Localizes newly added userscript details in `Greasemonkey` menu

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -97,6 +97,15 @@
   "matches": {
     "message": "Matches:"
   },
+  "menu_description": {
+    "message": "Description:"
+  },
+  "menu_homepage": {
+    "message": "Home page"
+  },
+  "menu_version": {
+    "message": "Version:"
+  },
   "new_user_script": {
     "message": "New user script..."
   },
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -15,6 +15,9 @@
   "cancel": {
     "message": "Cancel"
   },
+  "description": {
+    "message": "Description:"
+  },
   "disabled": {
     "message": "Disabled"
   },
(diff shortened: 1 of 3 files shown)


_locales/en/messages.json
```diff
@@ -97,6 +97,15 @@
   "matches": {
     "message": "Matches:"
   },
+  "menu_description": {
```
arantius: We don't anywhere else, and I don't want to get into the business of, trying to "categorize" the strings.

If the string is "Description", its name should be "description", and we should be able to use it anywhere we want to use that string.

Eselce: Okay!

Eselce: How will we differ labels and objects, i.e. for example ```Version:``` vs. ```version```?
I'd know the answer for ```Version $version$``` (that would be ```version_VERSION```)...

src/browser/monkey-menu.html
```diff
@@ -118,9 +118,9 @@
 
     <hr>
 
-    <p rv-if="activeScript.description">{'description'|i18n}
+    <p rv-if="activeScript.description"><strong>{'description'|i18n}</strong>
```
arantius: I'm not sure how you picked this HTML style, but I'd greatly prefer:

<p rv-if="activeScript.description">
  <strong>{'description'|i18n}</strong> {activeScript.description}
</p>

(Maybe a linebreak before `{` if that's >80 chars.

arantius: No, it was strong (copy/paste and edit of what was there), either is fine.  Bah markdown for eating (some) HTML tags.

Eselce: Your decision: &lt;b&gt; or &lt;strong&gt;, both are okay. I've started with &lt;u&gt; (which was really okay, but one should probably avoid underline here). The install dialog used &lt;strong&gt;. Current implementation is &lt;b&gt;.

PR #2937, opened 2018-04-06
Store blobs as base64 on backup

Sxderp:
Blobs JSON.stringify to empty objects. Base64 is used in order to
retain information.

### DIFF
--- a/src/bg/export-db.js
+++ b/src/bg/export-db.js
@@ -4,6 +4,20 @@
 // Private implementation.
 (function() {
 
+function blobToBase64(blob) {
+  if (!blob) return Promise.resolve(null);
+
+  let reader = new FileReader();
+  reader.readAsDataURL(blob);
+  return new Promise((resolve, reject) => {
+    reader.onload = event => {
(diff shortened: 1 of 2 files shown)


src/bg/export-db.js
```diff
@@ -4,6 +4,20 @@
 // Private implementation.
 (function() {
 
+function blobToBase64(blob) {
+  if (!blob) return Promise.resolve(null);
+
+  let reader = new FileReader();
+  reader.readAsDataURL(blob);
+  return new Promise((resolve, reject) => {
+    reader.onload = event => {
```
Eselce: Timing?

Sxderp:
I'm not sure what you mean. The closures passed to new promises are executed synchronously.

Eselce:
I am still not sure, if it is certain (or even needed), that `reader.onload` is set before `reader.readAsDataURL` is triggering some action (immediately started, but takes time) that results in using the `load` event... Maybe it's not critical, I just objected the possibility!

PR #2938, opened 2018-04-07
Resize menu on view switch

### LINKED ISSUE #2927: Long description causes version to flow off Monkey Menu
See picture, before and after rework.

before-after

Sxderp:
fixes #2927

### DIFF
--- a/src/browser/monkey-menu.css
+++ b/src/browser/monkey-menu.css
@@ -13,7 +13,7 @@ body {
   width: 300px;
   overflow: hidden;
 }
-body#rendering { display: none; }
+body.rendering { display: none; }
 
 
 #script-list-scroll {
@@ -36,13 +36,13 @@ section {
   width: 100vw;
 }
 
-section.user-script {
+section#user-script {
   position: absolute;
   top: 0;
   left: 100vw;
 }
 
-body#user-script section { margin-left: -100vw; }
+body.user-script section { margin-left: -100vw; }
 
 /********************************** MENUS ************************************/
 
@@ -127,7 +127,7 @@ section header {
 }
 
 
-section.user-script p {
+section#user-script p {
   margin: 0;
   padding: 4px 12px 4px 10px;
 }
--- a/src/browser/monkey-menu.html
+++ b/src/browser/monkey-menu.html
@@ -7,9 +7,9 @@
 <link rel="stylesheet" href="monkey-menu.css">
 <meta name="viewport" content="width=300, initial-scale=1">
 </head>
-<body id="rendering">
+<body rv-class="activeView">
 
-<section class="main-menu">
+<section id="main-menu">
   <menu>
     <menuitem id="toggle-global-enabled" tabindex="0">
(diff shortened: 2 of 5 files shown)


Eselce:
`rivets` error?

Sxderp:
Hm, most of the tests in the `monkey-menu.test.js` are just bad test. Unit tests aren't meant to 'check for syntax errors.' That's the job of a linter that should be run before the tests. There's one okay test, 'creates a tab for openUserScriptEditor()`, but even that's a bit iffy on what's actually being tested.

Even resolving the rivets issue another issue pops up because the DOM doesn't contain the elements that are known to exist in the menu. Since the tests aren't actually testing anything, the best option is to remove them and verify syntax through proper means.

Eselce:
As right as you are, wasn't the problem the transfer to the global scope, so that the "static" initialization didn't work? I thought, the (now removed) tests worked before (with `rivets` in `onload`)...

Sxderp:
> As right as you are, wasn't the problem the transfer to the global scope, so that the "static" initialization didn't work?

After fixing rivets, another problem came up. When the test calls [`uninstall`](https://github.com/greasemonkey/greasemonkey/blob/e3bddb3320be2898e57c0ee9c8ac190f50090d63/test/browser/monkey-menu.test.js#L17) iit goes through [`navigateToMainMenu`](https://github.com/greasemonkey/greasemonkey/blob/e3bddb3320be2898e57c0ee9c8ac190f50090d63/src/browser/monkey-menu.js#L268) which, in my PR, calls [`document.getElementById`](https://github.com/greasemonkey/greasemonkey/pull/2938/files#diff-6ba827f37e1348164a6c4d6c5d015654R200). Unfortunately, in the test environment this function call will return `null`. Since the element it is looking for does not exist, but it's known to exist in the real menu.

Sxderp:
Forgot to mention, the other tests also 'worked' because the test environment never calls `onLoad`. If it did, the rivets error would have occurred.

Eselce:
> 'worked' because the test environment never calls `onLoad`.

That's what I expected. Let's say, the tests were not thought through...

arantius:
The tests caught real errors that would have caused real pain for users.

If you remove them and don't add something better, that's a good way to get your branch not merged.

Sxderp:
I've already said what's better, a linter. If these tests (that don't actually test anything) have caught 'real errors' then I might want to reevaluate my environment. Assuming valid input into the functions (which the 'tests' do) anything causing these tests to be fatal would likely be caught by an IDE or manual linter, as all the tests do is check if the file contains valid JavaScript, not that the JavaScript is acting properly. 

Now, I assume most people have one of these on hand. Which is why I didn't propose anything. If you really want something in the repo I can code something with gulp that'll run some lint on all JS files and add it to the repo. But I thought that was overkill. 

Another alternative is to make it so these functions are actually tested. Unfortunately I'm not sure where to begin as many of the functions are intrinsically closely linked to the DOM. Research needs to be done on how to properly test UI linked code.

Issue #2877, opened 2018-02-26
Please display current version in the install dialog, if already installed


zuckel:
The install dialog currently gives no indication whether the user is installing a new script or updating an existing one. Please consider showing the currently installed version in case an existing script is going to be overwritten.

arantius:
A good UI mockup would make it easier to proceed.  I'm not sure I know how I'd want it to look.

Sxderp:
`oldversion` -> `newversion`

zuckel:
I'm with @Sxderp here. Simple does it.

Additionally I would, in case of an update, change the text on the install button to make the distinction clearer. If no version of the script is currently installed, it should still say `Install`. But if some existing version is going to be replaced, it should say something else, e.g. `Update`.

zuckel:
I'm looking into the code right now to see if I can create a pull request, but I am unfamiliar with the technologies used. Currently I am puzzled by two simple things:
- How do the properties like `version` find their way into `install-dialog.html`?
- What would be the correct way to make the text change? It seems that i18n happens late, so I would guess I would have to use a branch directly in `install-dialog.html`.

zuckel:
Alright, I've found my way around a little. Not sure when I find the time to really dig into this though. It seems the biggest change would to be getting the previous version of the script as the database is currently not consulted at all.

zuckel:
So, I've played around a little and I actually arrived at a minimal working product. It's ugly but I'm glad I got it working. Leaving this here for now as I'm heading away from my computer for today.

- Is it necessary to add translations for all known locales?
- Is `gDetails` actually necessary? It seems like a relict to me.
- The new Promise can certainly be cleaned up, though I don't know how.
- No idea if error handling is working/sufficient or not.
- Currently iterating over all installed scripts. A direct id lookup could be provided by the registry instead.
- There is no `id` getter on `scriptDetails`, therefore I am comparing `namespace` and `name` individually. There are probably some cases that are handled incorrectly here.

Sxderp:
> Is it necessary to add translations for all known locales?

No, just add English. Third-parties will help us with the rest.

> Is `gDetails` actually necessary? It seems like a relict to me.

I don't think so. But make a separate PR to fix this, rather than bundling it in yours.

> The new Promise can certainly be cleaned up, though I don't know how.

[...]

PR #2986, opened 2018-07-18
display if overwriting already installed script in install dialog #2877

zuckel:
I tried incorporating Sxderp's comments from #2877.

There is a new message UserScriptGetByName in the registry. It accepts namespace and name. There is no fuzzy matching. Only works if both are provided.

### DIFF
--- a/doc/Messages.md
+++ b/doc/Messages.md
@@ -127,6 +127,21 @@ Passes the user's option values from content to background for persistence.  Req
 
 * `excludes` A string, one `@exclude` pattern per line.
 
+# UserScriptGetByName
+Sent by: `content/install-dialog.js`
+Received by: `bg/user-script-registry.js`
+
+Returns installed user script by namespace and name.
+
+Data:
+
+* `scriptNamespace` The user script's namespace
+* `scriptName` The user script's name
+
+Response:
+
+* The `.details` object of the installed `RunnableUserScript`, or `false` if no user script was found.
+
 # UserScriptGet
 Sent by: `content/edit-user-script.js`
 
--- a/src/bg/user-script-registry.js
+++ b/src/bg/user-script-registry.js
@@ -161,6 +161,25 @@ function onListUserScripts(message, sender, sendResponse) {
 }
 window.onListUserScripts = onListUserScripts;
 
+function onUserScriptGetByName(message, sender, sendResponse) {
+  if (!message.scriptNamespace) {
+    console.warn('UserScriptGetByName handler got no namespace');
+  } else if (!message.scriptName) {
+    console.warn('UserScriptGetByName handler got no name');
+  } else {
+    let userScriptIterator = UserScriptRegistry.scriptsToRunAt(null, true);
+    for (let userScript of userScriptIterator) {
+      if (userScript.namespace === message.scriptNamespace
+          && userScript.name === message.scriptName) {
+        sendResponse(userScript.details);
+        return;
+      }
+    }
+    sendResponse(false);
+  }
+}
+window.onUserScriptGetByName = onUserScriptGetByName;
+
 
 function onUserScriptGet(message, sender, sendResponse) {
   if (!message.uuid) {
--- a/src/content/install-dialog.html
+++ b/src/content/install-dialog.html
@@ -22,7 +22,7 @@
     <tr>
       <td class="icon" rowspan="2"><img rv-src="iconUrl" width="40"></td>
       <td class="name">{name}</td>
-      <td class="version">{version}</td>
+      <td class="version"><span rv-if="installedVersion">{installedVersion} &rarr; </span>{version}</td>
     </tr>
     <tr>
       <td colspan="2">{description}</td>
@@ -101,7 +101,7 @@
   &nbsp;
   <button id="btn-install" class="button-big button-success"
       disabled="disabled"
-      >{'install'|i18n}</button>
+      ><span rv-if="installedVersion | not">{'install'|i18n}</span><span rv-if="installedVersion">{'update'|i18n}</span></button>
 </div>
 
 <script src="/src/i18n.js"></script>
--- a/src/content/install-dialog.js
+++ b/src/content/install-dialog.js
@@ -33,6 +33,24 @@ gDownloader.addProgressListener(() => {
 /****************************** DETAIL DISPLAY *******************************/
 
 gDownloader.scriptDetails.then(scriptDetails => {
+  return new Promise((resolve, reject) => {
+    chrome.runtime.sendMessage({
+        'name': 'UserScriptGetByName',
+        'scriptNamespace' : scriptDetails.namespace,
+        'scriptName' : scriptDetails.name
+      }, (installedScriptDetails) => {
+        if (chrome.runtime.lastError) {
+          console.error(chrome.runtime.lastError);
+          reject(chrome.runtime.lastError);
+        } else {
+          if (installedScriptDetails) {
+            scriptDetails.installedVersion = installedScriptDetails.version;
+          }
+          resolve(scriptDetails);
+        }
+    });
+  });
+}).then(scriptDetails => {
   gDetails = scriptDetails;
 
   document.title = _('NAME_greasemonkey_user_script', gDetails.name);
--- a/test/scripts/update.to.user.js
+++ b/test/scripts/update.to.user.js
@@ -1,0 +1,8 @@
+'use strict';
+// ==UserScript==
+// @name        Update test script 2 of 2
+// @namespace   test-namespace
+// @description Exercises the existing script update logic. Should show the old and the new version in the install dialog.
+// @grant       none
+// @version     2.0
+// ==/UserScript==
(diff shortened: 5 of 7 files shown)


Sxderp:
I'm current on my phone (so a lot of wrapped lines) but it looks okay. 

Small nitpick. In the registry lines 167 and 171 can be changed to `return Promise.resolve(userScript.details)` and `return Promise.resolve(false)`. 

Saves a grand total of one line. :no_mouth:

Sxderp:
Ah just saw something else. In the install dialog I believe you can do `rv-if="installedVersion"` and then `rv-else`. Rather than two if statements. 

Further you could probably put the conditionals on the button rather than using spans. But it's hard to tell on my phone due to wrapping. I'll have to do another review on a computer.

zuckel:
All other methods in the registry use the `sendResponse` parameter and the result is never returned directly. For the sake of consistency I would like to keep that.

I was trying to avoid duplicating the whole button. If there's a way to avoid duplication as well as the span, then of course we should change that. I just don't realize how.

The doc doesn't mention an `rv-else` and I can't make it work either. Maybe that's only in rivets?

I pushed a tiny commit changing the arrow to a nicer entity.

OT: There's an HTML comment closing with an extra dash in line 16, which kills formatting in VSC. Can I remove it or does it have some purpose?

Sxderp:
> The doc doesn't mention an `rv-else` and I can't make it work either. Maybe that's only in rivets?

I believe I was mistaken. Doesn't look like it exists.

> I was trying to avoid duplicating the whole button. If there's a way to avoid duplication as well as the span, then of course we should change that. I just don't realize how.

After some thought, if you line break the `<spans>` I think it might look overall cleaner than attempting to duplicate the button. Ex.
```html
<button id="btn-install" class="button-big button-success" disabled="disabled>
    <span />
    <span />
</button>
```

I haven't actually tested this patch yet, but to help facilitate testing you should add a script (or two) into the `test/scripts` directory which have the same name but different version numbers. Furthermore, it looks like you can 'update' to an older version. I wouldn't call that a bad thing, as some people may want to downgrade, just something to think about.

src/bg/user-script-registry.js
```diff
@@ -155,6 +155,24 @@ function onListUserScripts(message, sender, sendResponse) {
 }
 window.onListUserScripts = onListUserScripts;
```
Sxderp: nitpick: Two lines between functions

src/content/install-dialog.js
```diff
@@ -33,6 +33,25 @@ gDownloader.addProgressListener(() => {
...
+
+  return new Promise((resolve, reject) => {
+    chrome.runtime.sendMessage({
+        'name': 'UserScriptGetByName',
+        'scriptNamespace' : scriptDetails.namespace,
+        'scriptName' : scriptDetails.name
+      }, (installedScriptDetails) => {
+        if (chrome.runtime.lastError) {
+          console.error(chrome.runtime.lastError);
+          reject(chrome.runtime.lastError);
```
Sxderp: This reject is never going to get caught. There's no `catch` in the promise chain. Also, depending on the error it may be considered non-fatal and thus there's no need to stop the entire chain.

What you can do instead is remove line 44 and then on line 53 (number 52 after removing 44) add a `.catch` that logs the error, performs any additional checks, such as if the error is fatal or not, and if the error is not fatal then `return scriptDetails` or if it is fatal `throw [new]error`.

zuckel: How do I know if the error is fatal?

Sxderp: Good question. And for that I don't have a good answer. For this particular case the majority of what you might receive will likely be not fatal. Think of it this way, "If we fail to get the current version of a script is that a roadblock for displaying the rest of the install dialog?" For most situations I'd say no. However, if there are errors that we can handle or otherwise should display to the user (usually these are errors that we generate) then they should be handled appropriately before `return scriptDetails`.

Unfortunately it can be a major guessing game trying to figure out what Mozilla's APIs may throw at us and whether we can handle them or not.

src/bg/user-script-registry.js
```diff
@@ -155,6 +155,24 @@ function onListUserScripts(message, sender, sendResponse) {
...
 
+function onUserScriptGetByName(message, sender, sendResponse) {
+  if(!message.scriptNamespace) {
+    console.warn('UserScriptFind handler got no namespace')
```
arantius: Trailing semicolons are not optional.

src/bg/user-script-registry.js
```diff
@@ -155,6 +155,24 @@ function onListUserScripts(message, sender, sendResponse) {
 }
 window.onListUserScripts = onListUserScripts;
 
+function onUserScriptGetByName(message, sender, sendResponse) {
+  if(!message.scriptNamespace) {
```
arantius: Please follow existing code style, consistently.  There should always be a space after an if.  80 column maximum., probably some other things I haven't seen yet.

https://greasemonkey.github.io/style.html

src/bg/user-script-registry.js
arantius: "UserScriptFind" is not the name of any handler

src/bg/user-script-registry.js
```diff
@@ -155,6 +155,24 @@ function onListUserScripts(message, sender, sendResponse) {
...
+    console.warn('UserScriptFind handler got no name')
+  } else {
+    let userScriptIterator = UserScriptRegistry.scriptsToRunAt(null, true);
+    for (let userScript of userScriptIterator) {
+      if(userScript.namespace === message.scriptNamespace && userScript.name === message.scriptName) {
+        sendResponse(userScript.details);
+        return;
+      }
+    }
+    sendResponse(false);
```
arantius: What response gets sent in the case of one of the "warn" blocks above?

src/content/install-dialog.js
```diff
@@ -33,6 +33,25 @@ gDownloader.addProgressListener(() => {
 /****************************** DETAIL DISPLAY *******************************/
 
 gDownloader.scriptDetails.then(scriptDetails => {
+
```
arantius: What is this blank line for?  (IMO it does not belong.)

src/bg/user-script-registry.js
zuckel: I have no idea. I just did this analogous to `onUserScriptGet`.

src/content/install-dialog.js
zuckel: How do you continue with .then after .catch? I can't get this to work.

zuckel:
Minor cleanup and added test scripts. Error handling in install-dialog.js still unsolved.

src/content/install-dialog.js
Sxderp: Here's a gist. The main point is where you attach the `.catch`. You need to attach it such that you still have a reference to `scriptDetails`.
https://gist.github.com/Sxderp/6d37f4e42a02ec4f7f8962f4ef5579f4

doc/Messages.md
```diff
@@ -127,6 +127,21 @@ Passes the user's option values from content to background for persistence.  Req
 
 * `excludes` A string, one `@exclude` pattern per line.
 
+# UserScriptGetByName
```
arantius: We call this (namespace+name) the "ID".

I think we should keep the existing UserScriptGet message, and let it accept either an ID or a UUID.

src/content/install-dialog.html
```diff
@@ -22,7 +22,7 @@
     <tr>
       <td class="icon" rowspan="2"><img rv-src="iconUrl" width="40"></td>
       <td class="name">{name}</td>
-      <td class="version">{version}</td>
+      <td class="version"><span rv-if="installedVersion">{installedVersion} &rarr; </span>{version}</td>
```
arantius: line length

src/content/install-dialog.html
```diff
@@ -101,7 +101,7 @@
   &nbsp;
   <button id="btn-install" class="button-big button-success"
       disabled="disabled"
-      >{'install'|i18n}</button>
+      ><span rv-if="installedVersion | not">{'install'|i18n}</span><span rv-if="installedVersion">{'update'|i18n}</span></button>
```
arantius: line length

test/scripts/update.to.user.js
```diff
@@ -0,0 +1,8 @@
+'use strict';
```
arantius: FYI I just added https://github.com/greasemonkey/greasemonkey/blob/master/test/scripts/script-server-update.py which might obsolete the need for this?

zuckel:
@arantius should we go forward with this in general? It has been an interesting lesson so far, but I would ultimately like to see it merged. Do you think it makes sense once there is auto-update functionality? Check the screenshot for what it currently looks like:

gm2877

src/content/install-dialog.html
arantius: See `i18nBool` in `rivets-formatters.js`.

zuckel:
Closing this as I don't have time to finish it properly.

PR #2992, opened 2018-08-22
 Throw meaningful error on forbidden API call 

LWChris:
(Inspired by the e-mail to the greasemonkey-users mailing list sent by Brent on August 21st, 2018, where he was confused that even the minimalistic example taken from the GM wiki documentation resulted in a `GM.xmlHttpRequest is not a function` error)

Instead of only registering an API function like `GM.listValues` when access was granted, now otherwise register a function that'll throw a meaningful error. This way a script calling `GM.listValues()` w/o declaring `@grant GM.listValues` doesn't produce `TypeError: GM.listValues is not a function` in the error console, but rather `Error: <script name> does not @grant method GM.listValues`.

### DIFF
--- a/src/bg/api-provider-source.js
+++ b/src/bg/api-provider-source.js
@@ -14,7 +14,8 @@ function apiProviderSource(userScript) {
   if (!grants || grants.length == 0
       || (grants.length == 1 && grants[0] == 'none')
   ) {
-    return '/* No grants, no APIs. */';
+    /* No grants, no APIs. */
+    grants = [];
   }
 
   let source = '(function() {\n';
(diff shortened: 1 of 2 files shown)


Cerberus-tm:
I don't know if this appropriate, but thank you! I remember it taking a long time for me to find my error when I struggled with this some years ago.

makyen:
If this is done, then a script can not programmatically determine that a specific API function has not been granted by using `typeof`. It would have to resort to using `try`/`catch`, which, IME, is not used as often for a test like this.

Currently, it's easy to programmatically determine that the API has not been granted in a JavaScript-normal manner, because if the API is granted then `typeof GM.xmlHttpRequest === 'function'` will be `true`. If it's not granted, then that expression will be `false`. I've definitely used something like that to determine if a feature is available in Greasemonkey, although not quite in the same situation. It's possible that testing for existence/having been granted would be necessary in some situations where the functionality was *possible* to grant, but not actually granted (e.g. cross userscript-manager incompatibilities which the script author solved by not requesting a `@grant` in the way that it's needed for one manager, or another).

While these new error messages might be beneficial for a small number of cases where the userscript author isn't familiar with JavaScript/userscripts, the error

    TypeError: GM.xmlHttpRequest is not a function

is a typical JavaScript error, which clearly indicates that the property isn't a function. While a bit of experience is helpful in tracking down why it's not a function, doing so is a normal part of debugging JavaScript (or any programming language).

Frankly, I don't see the need for this change. While I have no examples of scripts which rely on the non-existence of these functions when they haven't been granted, it's certainly possible that there are some that do exist (i.e. the current implementation of this *may* break some scripts).

#### Alternate implementation method
If this really is something that is desired, perhaps implement a [`Proxy`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) for the `GM` Object which both outputs a "not granted" notice to `console.error` (not throw an error) when a not-granted property is accessed and still returns `undefined`, so that `typeof` will work properly and the normal JavaScript `TypeError` will be generated when used as a function (in addition to the above error notice). The drawback of using a `Proxy` would be a very slight performance hit to all `GM` API calls.

LWChris:
Showing meaningful instead of misleading errors is key to make any technology accessible to beginners. The e-mail mentioned in the PR is the best example.

But not only beginners. I am a programmer of 11 years, and I've been using userscripts since 2009. I'm not writing them very often though. And I can definitely say that a two weeks ago I would still have been confused by this error message. I had checked my code for typos, whether the function was removed or renamed, a conflict with other scripts, uninstalling and re-installing the script... A missing grant statement could've easily been the last thing coming to my mind. Meanwhile not so much anymore of course, now that I've dealt with this issue to such an extend... but back then it most certainly was possible I could've spent at least 10 minutes until I would see my mistake.

> Frankly, while I can see that there might be some benefit, I don't see the benefit outweighing the potential cost, with the current implementation.

I am convinced the number of scripts that were at least once broken because of a forgotten `@grant` exceeds the number of scripts currently relying on this behaviour by at least one order of maginute. Therefore _I_ can easily see the benefit outweighing the potential cost. ;)

> It would have to resort to using `try/catch`

No. If determination of available APIs is necessary, I'd rather simply _offer_ that functionality and properly document it.

    if (GM.isApiAvailable('xmlHttpRequest')) {
        /* Use API */
    } else {
        /* Workaround */
    }

appears to me as being far more understandable code than

    if (typeof GM.xmlHttpRequest === 'function') {
        /* Use API */
    } else {
        /* Workaround */
    }

People who rely on the API being either functional or undefined will end up with their scripts broken. Upon investigation what broke them, they will probably check the error console and see an error message that'll tell them "Your script used an API that it has not been `@grant`ed access to."

They will think "yeah that's why I was testing, too bad that doesn't work any more", and then they'll probably go to Google, search for "greasemonkey check grant" and find `GM.isApiAvailable` in the GM-wiki.

LWChris:
I am currently tidying up my repositories by deleting stale forks. So, once and for all:

Are you interested in ever merging something like this, maybe together with a `GM.isApiAvailable('<name>')` function that I'd happily include as well - or do you want to keep it the way that the function is simply not defined at all when it wasn't granted?

Because if there's no chance of realistically ever being included, I might as well delete the fork and move on.

makyen:
@LWChris In my comment above, I suggested an alternative method of implementation, which, potentially, would satisfy both the desire for outputing something in the console that's more informative to the user and which maintains compatibility with most existing code that might rely on currently functional methods of programmatically determining the presence of a particular GM API?. You haven't mentioned that you were unable to get that to work, so I'm assuming that you didn't attempt it. Is there a reason you haven't looked into alternate solutions which satisfy both requirements?

LWChris:
For the sake of simplicity, I'm going to use xmlHttpRequest as an example.

Adding a proxy could be a way of adding a console output AND return undefined. What would the console output be though? Because as long as the proxy returns "undefined", this would still yield the error log entry:

"TypeError: GM.xmlHttpRequest is not a function"

I would personally still rather get rid of the "is not a function" error line completely, rather than adding something onto it. I suppose that can't be done though. So something else has to be logged that is inherently more descriptive and explains why that "is not a function" line will show up second. Any suggestions? How about:

"Error: <script name> tried to access GM.xmlHttpRequest without permission. The function will be undefined unless the script header contains the appropriate @grant."

makyen:
The text you propose seems a bit verbose, but wouldn't be too objectionable. Text along the lines of what you had in your first message here seems more concise.

LWChris:
Yes, but would it be _too_ concise? I'm not sure people would understand that the following output

```
Error: <script name> does not @grant method GM.xmlHttpRequest.
TypeError: GM.xmlHttpRequest is not a function
```

is actually **one** error, not two. That's why I feel there should be some hint in the first message as to why the "is not a function" message would appear. Something along the lines of "As long as you don't fix this, there will be other problems." What do you think?

Cerberus-tm:
```
Error: <script name> needs "@grant GM.xmlHttpRequest" in the userscript's header.
TypeError: GM.xmlHttpRequest is not a function
```

Perhaps this would be the most 'noob friendly'?

PR #3054, opened 2019-09-29
Use CM comments add-on

feliperuiz:
It was already present in the codebase, just not being used. This enables quick
line toggling by pressing `Ctrl-/` or `Cmd-/` on mac.


arantius:
Thanks!  I'm not sure when the next release will be, but this will be included.

PR #3058, opened 2019-10-23
Add accessibility semantics and labels.

jcsteh:
This makes Greasemonkey far more usable by screen reader users.

Even though keyboard navigation is implemented for the Monkey Menu, this change exposes it as buttons, rather than menu + menuitems. This is because unlike most menus, this menu contains headings, which doesn't fit the ARIA menu pattern. This way, screen reader users will be able to browse this like any other document and will be made aware of (and able to jump to) the headings.

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -258,5 +258,11 @@
   },
   "xhr_no_url": {
     "message": "GM.xmlHttpRequest: Received no URL."
+  },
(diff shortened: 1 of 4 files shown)


MarcoZehe:
Is there any way to move this forward? This PR has been sitting here uncommented and unreviewed for several months.

arantius:

PR #3059, opened 2019-10-23
Expose sections as dialogs for accessibility.

jcsteh:
Previously, screen readers reported nothing when the user switched to a new section. With this change, each section is exposed as a dialog, with the active dialog getting focus. This way, a screen reader will report the content of the new section as if a dialog appeared.

I separated this from #3058 because this one is a bit more controversial, since it messes with focus a tiny bit. Also, I'm totally blind, so I can't confirm whether focusing the sections causes problems visually.

### DIFF
--- a/src/browser/monkey-menu.css
+++ b/src/browser/monkey-menu.css
@@ -41,6 +41,14 @@ section {
   width: 100vw;
 }
 
+/* The active section gets focus so screen readers report the switch to a new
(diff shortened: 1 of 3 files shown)


MarcoZehe:
Any way to move this forward? This PR has been sitting open and unreviewed for several months.

arantius:

PR #3060, opened 2019-10-24
Add an optional simple editor which uses a plain textarea.

jcsteh:
This is needed by screen reader users, as CodeMirror is unfortunately not accessible. Having a separate editor is not ideal, nor is the missing functionality for screen reader users. Nevertheless, this is far better than a completely inaccessible editor. This at least allows screen reader users to read the content and do basic editing.

A more ideal alternative would be switching to [Monaco](https://microsoft.github.io/monaco-editor/), which is accessible. This would avoid the need for a separate editor and provide advanced functionality for all users. I'm not sure if that is feasible, though.

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -258,5 +258,14 @@
   },
   "xhr_no_url": {
     "message": "GM.xmlHttpRequest: Received no URL."
+  },
+  "editor": {
+    "message": "Editor"
+  },
+  "use_enhanced_editor": {
+    "message": "Use enhanced editor"
+  },
+  "use_enhanced_editor_explain": {
+    "message": "Screen reader users should disable this, as the enhanced code editor is not accessible."
   }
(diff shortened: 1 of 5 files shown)


jcsteh:
I discovered that this broke when creating a new script, as the ?simple query string wasn't passed in that case. Fixing that was going to introduce more ugliness, so I refactored so that the options are fetched in edit-user-script. That should also make it easier to add additional editor configurability in future.

arantius:
Monaco looks interesting and [this example](https://github.com/microsoft/monaco-editor-samples/blob/master/browser-script-editor/index.html) makes me think that if we include its files the right way, we can load it from a web extension.

MarcoZehe:
Any way to move this forward? This PR has been sitting here largely uncommented and unreviewed for several months.

arantius:

PR #3100, opened 2021-01-15
ion monkey crash fix

Soldy:
I had a random crash problem. 
I find a situation most of the simple function calls can be crash if the ion monkey is in use. The editor sometime still crashing.


Soldy:
Too much change :(

PR #3114, opened 2021-04-19
Disable spellcheck for simple editor

dogancelik:
spellcheck - no spellcheck-fs8

Here's a comparison.

### DIFF
--- a/src/content/edit-user-script.js
+++ b/src/content/edit-user-script.js
@@ -14,6 +14,7 @@ class SimpleEditor {
     this._onSwapDoc = null;
     this._textarea = document.createElement("textarea");
     this._textarea.style["white-space"] = "pre-wrap";
+    this._textarea.spellcheck = false;
     this._textarea.addEventListener("input", this._onChange.bind(this));
     this._textarea.addEventListener("keydown", this._onKeyDown.bind(this));
     element.appendChild(this._textarea);


Soldy:
can you check this ? 
[https://github.com/Soldy/greasemonkey/commit/86b1950d143339ea5be7c6001cb5d4a69ee1cfa4](https://github.com/Soldy/greasemonkey/commit/86b1950d143339ea5be7c6001cb5d4a69ee1cfa4)

makyen:
I disagree with disabling spellcheck here. Having it be enabled or disabled is really a personal preference. Personally, I generally find having spellcheck enabled beneficial, It's easy to just add a substantial number of the words which are indicated as misspelled to your dictionary, which you only have to do once, so they no longer are indicated as misspelled (in any of your userscripts). I'd argue that, based on the image in the original comment here, you aren't using this feature correctly.

I would support having it be something the user can choose to enable/disable.

However, if it's not reasonable/possible to have it be user-configurable, then it should remain enabled, IMO. Basically, if someone wants it and it's disabled, they can't get it. If someone doesn't want it and it's enabled, then they can just ignore the formatting (i.e., it's just annoying). So, the least harm is to have it enabled, if it's not possible to have it be user-configurable.

Overall, this change is just imposing a personal preference upon everyone. Admittedly, the fact that it's on is imposing the opposite preference, but, again, that's the least harm of the two non-user-configurable options, IMO.

Soldy:
> I disagree with disabling spellcheck here. Having it be enabled or disabled is really a personal preference. Personally, I generally find having spellcheck enabled beneficial, It's easy to just add a substantial number of the words which are indicated as misspelled to your dictionary, which you only have to do once, so they no longer are indicated as misspelled (in any of your userscripts). I'd argue that, based on the image in the original comment here, you aren't using this feature correctly.
> 
> I would support having it be something the user can choose to enable/disable.
> 
> However, if it's not reasonable/possible to have it be user-configurable, then it should remain enabled, IMO. Basically, if someone wants it and it's disabled, they can't get it. If someone doesn't want it and it's enabled, then they can just ignore the formatting (i.e., it's just annoying). So, the least harm is to have it enabled, if it's not possible to have it be user-configurable.
> 
> Overall, this change is just imposing a personal preference upon everyone. Admittedly, the fact that it's on is imposing the opposite preference, but, again, that's the least harm of the two non-user-configurable options, IMO.

Please note you probably haven't checked what I try to suggest. My suggestion is a simple freedom of choice. A menu point that lets you disable the spellcheck just for the grease monkey.

```javascript 
    this.#textarea.style["white-space"] = "pre-wrap";
    this.#textarea.spellcheck = options.simpleEditorSpellCheck;
    this.#textarea.addEventListener("input", this.#onChange.bind(this));
    this.#textarea.addEventListener("keydown", this.#onKeyDown.bind(this));
    this.#element.appendChild(this.#textarea);
```

 That's all. I do not go with the force disable.

PR #3143, opened 2022-06-18
Add feature syncing user scripts with local files via WebDAV server to be able to use external editor

### LINKED ISSUE #2513: WebExt: External editor support (read/write extension directory)
The [runtime](https://developer.chrome.com/extensions/runtime) API provides multiple [ways](https://stackoverflow.com/questions/17179807) to read files from the extension directory, so what is left to explore is how to write files to the extension directory.

I think it's worth exploring if [runtime#getPackageDirectoryEntry](https://developer.chrome.com/apps/runtime#method-getPackageDirectoryEntry) and [fileSystem#getWritableEntry](https://developer.chrome.com/apps/fileSystem#method-getWritableEntry) can used to write files to the extension directory, but I fear that the permission system might interfere.

### LINKED ISSUE #2932: Accessing and editing files with the help of NiklasGollenstede's native-ext
Even if as an extension to greasemonkey, are there any opinions or tests for making code to work in conjunction with [NiklasGollenstede's native-ext](https://github.com/NiklasGollenstede/native-ext) to provide back file editing in the user's preferred editor?

### LINKED ISSUE #3048: Not so hot auto reload
a) Allow the user to whitelist a development local domain or IP where you can add a script that will automatically check for updates every second (only while the tab is focused?), or
b) that you can at least bookmark its URL and reload it without having to wait and click on Install.

Auto reloading the page would get annoying while testing (and not the intended behavior for testing), so instead the user would be responsible to reload the page itself when he wants to.

For security, this script could only work temporarily (until you close the browser or require manual "reinstall" every N minutes). There could be a warning color in the Greasemonkey icon or in other places making it clear that it's running.

esperecyan:
This PR resolves https://github.com/greasemonkey/greasemonkey/issues/2513, resolves https://github.com/greasemonkey/greasemonkey/issues/2932, and resolves https://github.com/greasemonkey/greasemonkey/issues/3048.

- Make a PROPFIND request every second to detect file updates from etag and last-modified.
	+ Save etag and last-modified on Greasemonkey side.
- Save as `Script Name (UUID)/Script Name.user.js`.
	+ Convert characters that cannot be used in the file name to full-width ASCII and truncate if too long.
	+ Identify the script by the UUID in the folder name.
- Errors when `@require` or `@icon` fails to load, or when name and namespace is duplicated, are notified by `browser.notifications`.
	+ In case of errors, etag and last-modified on Greasemonkey side will be updated to avoid re-loading files with errors.

To make it easy to build a persistent local WebDAV server, I have prepared a template for Docker.
https://github.com/esperecyan/webdav-server-for-local-client

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -87,6 +87,9 @@
   "fix_and_save": {
     "message": "The editor contents have not been saved.  Fix this error and try again."
   },
+  "fix_and_save_sync_via_webdav": {
+    "message": "The modified user JS file have not been synced.  Fix this error and try again."
+  },
   "get_user_scripts": {
     "message": "Get user scripts"
   },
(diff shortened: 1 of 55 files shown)


arantius:
This does look interesting, but it's quite a lot to merge and support forever.  It would have been nice to discuss this before writing the code, I can imagine simpler ways to accomplish the same.

This isn't likely to be merged.

Cerberus-tm:
This looks like great work! Would be really cool to have built-in support for external editors like this.

esperecyan:
@arantius I believe that some level of complexity is inevitable due to the need to use browser.storage as the main, and optional synchronization with the file system. Even if create original server application instead of WebDAV and use Native messaging, I think that most of this code can be diverted.

@makyen That is 1 request per second covering every userscripts by PROPFIND method with `depth: Infinity`.

PR #3144, opened 2022-06-18
Commit package-lock.json

esperecyan:
I do not know why it is not committed, but I do not see any disadvantage.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,6065 @@
+{
+  "name": "Greasemonkey",
+  "version": "0.0.1",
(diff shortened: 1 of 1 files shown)


Soldy:
I am highly against this in open source. This can easily take over any developer's machine. Because the package lock file is much larger than the package.json, and no one has time to check the full list. A malicious package can easily be injected into the package lock file. 

In a close team where u can trust everyone, that should be fine. In open source, that may not be the best. 



And by the way, that already 3 years old :D

PR #3176, opened 2023-11-12
Tooltip for userscript menu command

iampradip:
Since menu command can overflow and display '...' for long command caption

### DIFF
--- a/src/browser/monkey-menu.html
+++ b/src/browser/monkey-menu.html
@@ -163,7 +163,7 @@
       tabindex="0"
       role="button">
     <i class="icon"><img rv-src="command.icon"></i>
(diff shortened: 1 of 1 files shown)


arantius:
This has the undesired side effect of adding a tooltip to every menu item, even the ones that are not truncated.  There's no obvious good way to do this only when needed.

Do you have a concrete example of a script that has menu commands with long names?

arantius:
Yes but which scripts are adding those commands?  Are they public scripts you can link to?

iampradip:
> Yes but which scripts are adding those commands? Are they public scripts you can link to?

No, they are all private.

Issue #3190, opened 2024-07-16
Automatic updates seem to never happen


binki:
I have a script which I updated on 2024-04 but one of my browsers says it last updated it on 2023-05-22. Automatic updates is on and everything.

When investigating, I looked at the code in the debugger. I think I found two issues.

# Lack of `updateNextAt` value

https://github.com/greasemonkey/greasemonkey/blob/c0b3b542074b48793fb24023da114cfeb643ec95/src/bg/updater.js#L104

I also found that only one of my scripts had a `updateNextAt.` value set in `chrome.storage.local.get()`.

It seems that `v < nextTime` will evaluate to `false` when `v` is `undefined` and `nextTime` is some positive number:

```
> undefined < 10
false
```

I think that the intention is that any script without a `updateNextAt.` value automatically be scheduled for an immediate update check. What the actual current logic appears to be is that any script which had ever successfully updated in the past will update. It seems impossible for more than one script to ever have an `updateNextAt.` value set for it unless if the user debugs the extension and manually adds keys. So I think that either `v === undefined` should be added to the list of short-circuiting conditions or v should be coalesced to `0` when its value is not present in `chrome.storage.local`.

To explain in more detail: it seems that on one of my systems, only one script gets updated due to this issue. On another system, only 7 of 14 or so eligible scripts gets updated. The reason that sometimes multiple scripts actually can get their `updateNextAt.` value set depends on the order that script UUIDs are added to the `updateNextAtKeys` variable. As long as a newly added script which does not have an `updateNextAt.` value yet is the first key in `updateNextAtKeys` during at least one auto-update check, then it will, as the author intended, be considered eligible for updating and join the list of scripts which are capable of auto-updating (due to successfully having its `updateNextAt.` value set during that update check).

# Lack of Error Tolerance

At https://github.com/greasemonkey/greasemonkey/blob/c0b3b542074b48793fb24023da114cfeb643ec95/src/bg/updater.js#L131 I see a call to a function which can throw (i.e., awaiting a `Promise` which can be rejected). In the above example, when checking the values on one of my computers, the one `updateNextAt.` value I saw in the system was set to a value in the past. Since this is running on a browser which might be left open for weeks on a computer which does sometimes enter a low power state, I suspect, but cannot prove, that the update checks of even that one script ended up paused due to updates never being rescheduled after a download failure. I think a `try{}finally{}` construct is in store here. This would enable the extension to continue checking for updates even if the scheduled update occurs when the system is offline.

[...]

PR #3191, opened 2024-07-17
Fix issue preventing some scripts from auto-updating.

binki:
GreaseMonkey computes the next scheduled update for each script each time it successfully automatically checks for an update. This information is used to select the next script which automatically check for an update. This change ensures that scripts which do not have this information stored (i.e., scripts which have yet to have a successful automatic update check) will perform their initial automatic update check regardless of the order in which scripts are iterated.

Due to the existing parameters and tuning in GreaseMonkey, this change will result in any update-eligible scripts which have not had their updates checked causing a 3-hour pause of updates. You must ensure that you keep your browser open for at least 3 hours for each script which has not had an automatic update check performed on it before other scripts can resume their normal update schedule.

See #3190 and the discussion of updateNextAt.

### DIFF
--- a/src/bg/updater.js
+++ b/src/bg/updater.js
@@ -95,20 +95,21 @@ window._pickNextScriptAutoUpdate = async function() {
       return;
     }
 
-    let defaultCheckTime = new Date().getTime() + MIN_UPDATE_IN_MS;
     chrome.storage.local.get(updateNextAtKeys, vs => {
       for (let k of updateNextAtKeys) {
         let uuid = k.replace('updateNextAt.', '');
-        let v = vs[k];
+        // The key will be absent until the script is checked for updates.
+        let v = vs[k] || 0;
 
-        if (!nextTime || !nextUuid || v < nextTime) {
-          nextTime = v || defaultCheckTime;
+        if (!nextUuid || v < nextTime) {
+          nextTime = v;
           nextUuid = uuid;
         }
       }
 
       if (nextUuid) {
-        resolve([nextUuid, nextTime]);
+        let defaultCheckTime = new Date().getTime() + MIN_UPDATE_IN_MS;
+        resolve([nextUuid, nextTime || defaultCheckTime]);
       } else {
         reject('Could not find next script to update.');
       }


src/bg/updater.js
```diff
@@ -101,7 +101,8 @@ window._pickNextScriptAutoUpdate = async function() {
         let uuid = k.replace('updateNextAt.', '');
         let v = vs[k];
```
arantius, line 102: Thanks so much for your effort!

After reading all the discussion and thinking about it, what do you think of making this line `let v = vs[k] || 0;`?  That's probably self-documenting.  And would still fix the underlying problem?

binki: Yes, I will do that and just remove the `defaultCheckTime` variable altogether. I think that would provide a much better experience for users.

binki: Or, actually, I do want to clarify. Are you OK with removing `defaultCheckTime`? This way, any extension which has not been updated yet will be updated immediately (well, after 10 seconds ( https://github.com/greasemonkey/greasemonkey/blob/b0f597970cd7205458675a83c2b6783ac34f9925/src/bg/updater.js#L129 )). This removes the requirement that the browser be running for 3 hours (which might especially be a stretch for Android) for any scripts to update when at least one has not been updated before.

binki:
Note that since `v` can be `0`, that means that `nextTime` can be `null` or `0` and the condition has to distinguish those. Thus the check for `nextTime === null`.

src/bg/updater.js
makyen: 10 seconds after Firefox startup seems too short for this activity. In the past, there's been great concern from Mozilla about how long it takes Firefox to start up. I'd suggest a minimum value of at least several minutes (10? 15?) to assure that this additional processing, which isn't actually required for startup, is pushed well outside of the time during which the user is wanting to immediately interact with Firefox after startup.

binki: > 10 seconds after Firefox startup seems too short for this activity. In the past, there's been great concern from Mozilla about how long it takes Firefox to start up. I'd suggest a minimum value of at least several minutes (10? 15?) to assure that this additional processing, which isn't actually required for startup, is pushed well outside of the time during which the user is wanting to immediately interact with Firefox after startup.

Don’t worry. There already is a 15-minute pause because the first attempt at updates happens before the script registry loads, throwing an error which is caught at which point the first real automatic check is scheduled (I would like to make another separate change so that this is more explicit instead of accidental-looking—if I have the time):

https://github.com/greasemonkey/greasemonkey/blob/b0f597970cd7205458675a83c2b6783ac34f9925/src/bg/updater.js#L124

The part of logic that I am changing is recording the next attempt time on a script-by-script basis. If Firefox has been unopened for months, all scripts will have a next update time in the past. This use of the value of 0 for a never-before-update-checked script is identical to that scenario of scripts which have not been updated for months due to the browser not running for months.

binki: @makyen though, while this is off-topic for this particular PR, I personally would prefer that scripts which haven’t been updated for weeks are immediately updated upon opening Firefox. Since most scripts are updated due to website compatibility, staying out of date will provide a worse experience than any slowdown due to checking. Also, GreaseMonkey already serializes updates, so there is a limit to the performance impact.

makyen: @binki Saying things like "this is off-topic, but I'm going to state my position anyway" is, generally, unhelpful to keeping a conversation on-topic, despite the fact that we all do it (including me), at least on occasion. By doing so, the other person is put in the position of either not responding and allowing the already expressed position be the only one stated there, which, at least to some readers, will be seen as implicit agreement, or to respond substantially to the off-topic comment and continue the off-topic conversation (which may or may not be appropriate, and likely disruptive to those who have chosen to get notifications of new content for whatever was on-topic).

"Should non-required things be done at startup or should they be delayed to later in order to make startup faster" is something that has had quite a bit of discussion over the years, particularly with respect to Firefox. Mozilla has come down *strongly* directing that everything that's not actually required should be delayed past startup, as Mozilla sees customer feedback of long startup times as one of the significant reasons users switch away from using Firefox.

While delaying non-required tasks is the correct choice in general, and it is something that extension writers should be significantly considering while writing their extension, if it's the right choice in any specific instance, is, often, debatable, and likely something that individual users would have differing opinions on.

binki: @makyen There, I opened #3193 for you.

src/bg/updater.js
```diff
@@ -95,14 +95,13 @@ window._pickNextScriptAutoUpdate = async function() {
...
     }
 
-    let defaultCheckTime = new Date().getTime() + MIN_UPDATE_IN_MS;
     chrome.storage.local.get(updateNextAtKeys, vs => {
       for (let k of updateNextAtKeys) {
         let uuid = k.replace('updateNextAt.', '');
-        let v = vs[k];
+        let v = vs[k] || 0;
 
-        if (!nextTime || !nextUuid || v < nextTime) {
```
arantius, line 104: Please keep code consistent with the existing style (keep `!nextTime`).

arantius:
Keep changes small please.  If you want to fix a bug, fix it.  If you want to change a behavior, argue for/against that in its own PR.  Thanks.

src/bg/updater.js
binki: @arantius You had said earlier that you wanted to use `let v = vs[k] || 0;`, so I took that to be condoning `nextTime !== null`. I'll have to do `let v = vs[k] || defaultCheckTime;` to actually solve the problem. Not sure if that will meet your style criteria, but I will prepare that variant of the change and restore the caveat of the 3 hour minimum browser runtime to the commit message.

binki: Err, even that doesn't work. I'll have to use a value of 1 there and check that inside of the `if`. This is getting way more complicated than removing the odd 3 hour pause behavior.

binki:
The bug is a behavioral one. I conflated an intentional behavior with the behavior that results in the bug.

The changes have been updated to ensure:

* The `_pickNextScriptAutoUpdate()` function prioritizes returning a script which does not have the `updateNextAt.` key set for it yet over any other script. This is done by setting `v` to `0` when `vs[k]` is `undefined` which enables the `v < nextTime` check to be `true` when `nextTime` is a value loaded from a script which does have the `updateNextAt.` key set for it. This is the `let v = vs[k] || 0;` behavior requested by https://github.com/greasemonkey/greasemonkey/pull/3191#discussion_r1682829354 .
* The `_pickNextScriptAutoUpdate()`, per your request at https://github.com/greasemonkey/greasemonkey/pull/3191#issuecomment-2237751594 , preserves the behavior of waiting `MIN_UPDATE_IN_MS` for any script which does not have the `updateNextAt.` key set for it. This is unfortunately a floating value (so restarting the browser every 3 hours and 12 minutes will prevent automatic updates from ever proceeding due to this behavior, but that will get its own issue tracker).

PR #3206, opened 2025-02-22
Fix XHR ignoring Cookie header unless it is last

### LINKED ISSUE #3205: hasCookieHeader works only when Cookie header is set as the last property of headers
`hasCookieHeader` in [`on-user-script-xhr.js`](https://github.com/greasemonkey/greasemonkey/blob/master/src/bg/on-user-script-xhr.js#L110) hold the boolean value indicates whether the `headers` contains `Cookie`.

However, this variable holds `true` only if the `Cookie` is the last property of headers because the `hascookieHeader` will be overwritten with the new value for each loop.

## Solution
In order to solve this problem, we can change the code like this:

```patch
  let hasCookieHeader = false;
  if (d.headers) {
    for (let prop in d.headers) {
      if (Object.prototype.hasOwnProperty.call(d.headers, prop)) {
        let propLower = prop.toLowerCase();
-        hasCookieHeader = (propLower === 'cookie');
+       hasCookieHeader ||= (propLower === 'cookie');
        if (gHeadersToReplace.includes(propLower)) {
          xhr.setRequestHeader(gDummyHeaderPrefix + propLower, d.headers[prop]);
        }
        else {
          xhr.setRequestHeader(prop, d.headers[prop]);
        }
      }
    }
  }
```

syusui-s:
Resolves #3205

The current implementation of XHR ignores `Cookie` header because `hasCookieHeader` is not updated correctly.
This PR fixes the issue.

### DIFF
--- a/src/bg/on-user-script-xhr.js
+++ b/src/bg/on-user-script-xhr.js
@@ -112,7 +112,7 @@ function open(xhr, d, port, tabUrl) {
     for (let prop in d.headers) {
       if (Object.prototype.hasOwnProperty.call(d.headers, prop)) {
         let propLower = prop.toLowerCase();
-        hasCookieHeader = (propLower === 'cookie');
+        hasCookieHeader ||= (propLower === 'cookie');
         if (gHeadersToReplace.includes(propLower)) {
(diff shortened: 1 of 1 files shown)


Cerberus-tm:
Thank you!

Issue #3117, opened 2021-05-27
Script doesn't run when using `firefox.exe <url>`


maccadia:
Hi,

With this simple script : 
```js
// ==UserScript==
// @name     Anonyme Script 387564
// @match		 https://www.dokuwiki.org/*
// @version  1
// @grant    none
// ==/UserScript==

console.log("test");
```

The script does not always run when I start firefox from command line like this : 

On Linux `firefox "https://www.dokuwiki.org/dokuwiki?do=login&sectok="`

On Windows ` & 'C:\Program Files\Mozilla Firefox\firefox.exe' "https://www.dokuwiki.org/dokuwiki?do=login&sectok="`

The same behavior happens when the URL is the start page for Firefox.

When I reload the page, the script runs.

I tested with ViolentMonkey and it doesn't seem affected.

afontenot:
Could not replicate on Linux with Firefox 89. Tried 5 times from a cold start. Possibly some kind of race condition?

xpxaxsxi:
This bug is very much alive, the following script does not run in Windows (tested on Windows 11)
--when using a url with  firefox.exe in command line
--when Firefox has setting open earlier windows ON

Refreshing a such page always works 

Fresh Firefox profiles and adding GM to that, that does not help
```
// ==UserScript==
// @name     Anonyme Script 387564
// @match		 https://www.wikipedia.org/*
// @version  1
// @grant    none
// ==/UserScript==

console.log("test");
```

Cerberus-tm:
This would explain why I regularly find tabs where scripts do not seem to have loaded. I wonder whether Greasemonkey could fix this, or whether it is a Firefox issue?

binki:
> Maybe there’s some event it could hook into so that it can prevent pages from loading before it is ready?

Actually, this is a bad idea since it would impact performance. Instead, Greasemonkey should apply itself to already-loaded tabs once it is loaded if there is a race condition between loading initializing extensions and loading tabs/restoring session. (I.e., does Firefox guarantee that `webNavigation.onCommitted` for session restore will be observed by extensions which are installed prior to Firefox being launched or not? If not, then Greasemonkey has to somehow compensate).

[...]

PR #3207, opened 2025-02-24
Fix race condition with data load when checking for scripts to run.

binki:
This addresses one race condition discussed in #3117 where scripts would not be run on a tab if the webNavigation.onCommitted event arrives prior to the list of available scripts fully loading.

### DIFF
--- a/src/bg/execute.js
+++ b/src/bg/execute.js
@@ -6,9 +6,10 @@ content script executions.
 TODO: Make document_start execution time work as intended.
 */
 
-function executeUserscriptOnNavigation(detail) {
+async function executeUserscriptOnNavigation(detail) {
   if (false === getGlobalEnabled()) return;
 
+  await userScriptsReady;
   const userScriptIterator = UserScriptRegistry.scriptsToRunAt(detail.url);
   for (let userScript of userScriptIterator) {
     let options = {


src/bg/execute.js
```diff
@@ -6,9 +6,10 @@ content script executions.
 TODO: Make document_start execution time work as intended.
 */
 
-function executeUserscriptOnNavigation(detail) {
+async function executeUserscriptOnNavigation(detail) {
```
arantius, line 9: Oh of course, need async to await ...

But A) the test that calls this should await this now, and B) I can't so far find any documentation either way: does `webNavigation.onCommitted.addListener()` accept an async function?  I'd be more comfortable if I knew it was documented to do so (or to not...).

Sxderp: As far as I'm aware all WebExtension `addListener`s accept async functions. I can only think of one instance where an async function is not desirable. The `onMessage` handler. 

https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/onMessage#addlistener_syntax

arantius: Yes that doc talks about async functions, and about returning promises.  So .. it's the exception that proves the rule?

Or webNavigation doesn't use the listener's return value, so doesn't care?

binki: I am pretty sure that `webNavigation.oncommitted` is a fire-and-forget event. The [API documentation](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/webNavigation/onCommitted#listener) doesn’t describe the handling of the return value. I considered it harmless to return a `Promise`. This is different than, say, [`webRequest.onHeadersReceived`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/webRequest/onHeadersReceived#listener) which expects a response and is explicitly documented as supporting a `Promise` on Firefox (but apparently not on Chrome). I did consider whether or not returning a `Promise` would be damaging for this event, but I expected it not to matter and in testing with the browser I never encountered any issue.

What I was concerned about when authoring this was whether or not introducing asynchronous behavior in the event would be a problem. I noticed that the event **already is asynchronous** when it comes to actual script execution and error logging: https://github.com/greasemonkey/greasemonkey/blob/8a51d1432e8c1adbd761d2654b030e11efc11b74/src/bg/execute.js#L20 , so I dismissed that concern.

Speaking of testing, I am not sure if the test should `await` that API. The reason is what I just stated above: the function is already asynchronous and, depending on how you are actually performing the test or what you are testing, the effects won’t be synchronously observable anyway. For example, verifying that the error logging works with a mock that actually behaves like the real browser would require handling asynchronous side-effects of the function.

I did not realize that the CI did not run the tests. I thought that CI being all green meant they were passing, but I now see they are not. I see now that my changes do cause the test to fail because the test uses mocks and assumes that the function runs synchronously. I guess that, for the existing test, the easiest thing to do would be to make it await the `Promise`. But the test could be updated to handle asynchronous execution. Then it would pass even if we decide to not have the handler return a `Promise`.

However, I would like to take a step back and recognize that I wasn’t thinking in terms of testing at all. What I would like to ideally do is introduce a test which fails without this change. This could be used to detect and avoid regressions.

arantius: Overall this is a great response, thanks!

There is no CI, anymore, not intentionally.  Where did you see that it was green?  (There used to be, but Travis broke it, and development is so slow now, it wasn't worth the effort to fix it.)

PR #3215, opened 2025-07-24
Github CI implementation. 

Soldy:
This commits likely can resolves the Tavis and Circle CI issues as well, but I am not entirely sure about that. 
I wish to find another solution than that. But I was made a costume chai.js. Because the chai breaks with karma, do a stupid export. 
Maybe the step away from chai is the best solution in that case.
However, this fixes the test for now.

### DIFF
--- a/.github/workflows/karma.js.yml
+++ b/.github/workflows/karma.js.yml
@@ -1,0 +1,26 @@
+
+name: Node.js CI
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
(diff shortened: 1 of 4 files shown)


Soldy:
The problem : 


```

Firefox 140.0 (Linux x86_64) ERROR 


 SyntaxError: export declarations may only appear at top level of a module 


 at node_modules/chai/chai.js:4151:1


```

I will try to find another solution than costume  chai.js, but I only had 2 easy options. Go with older nodejs, or this.

PR #3216, opened 2025-07-25
Github CI implementation. 

Soldy:
This commits likely can resolves the Tavis and Circle CI issues as well, but I am not entirely sure about that.
I wish to find another solution than that. But I was made a costume chai.js. Because the chai breaks with karma, do a stupid export.
Maybe the step away from chai is the best solution in that case.
However, this fixes the test for now.

The problem : 

```

Firefox 140.0 (Linux x86_64) ERROR 


 SyntaxError: export declarations may only appear at top level of a module 


 at node_modules/chai/chai.js:4151:1


```

I will try to find another solution than costume  chai.js, but I only had 2 easy options. Go with older nodejs, or this.

### DIFF
--- a/.github/workflows/karma.js.yml
+++ b/.github/workflows/karma.js.yml
@@ -1,0 +1,26 @@
+
+name: Node.js CI
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [23.x]
+
+    steps:
+    - uses: actions/checkout@v4
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v4
+      with:
+        node-version: ${{ matrix.node-version }}
+    - run: npm i --verbose --dev
+    - run: npm ls -a 
+    - run: npm test
--- a/chai.js
+++ b/chai.js
@@ -1,0 +1,4134 @@
+var __defProp = Object.defineProperty;
+var __getOwnPropNames = Object.getOwnPropertyNames;
+var __name = (target, value) => __defProp(target, "name", { value, configurable: true });
+var __commonJS = (cb, mod) => function __require() {
+  return mod || (0, cb[__getOwnPropNames(cb)[0]])((mod = { exports: {} }).exports, mod), mod.exports;
+};
+var __export = (target, all) => {
+  for (var name in all)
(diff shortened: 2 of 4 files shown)


arantius:
First, thank you for your interest and effort.

That said: the change rate of this repository is so low anymore that I don't consider CI a high priority.  I definitely will not merge anything that doesn't A) put third party code in the right location (see existing examples) and B) completely and correctly comply with any third party code's licenses.

So this should probably simply be closed "wontfix".

Plus, please spell "custom" correctly.

Soldy:
The chai software is licensed under the MIT License, and I don't believe that use of it violates any copyright, especially since this is not a case of redistribution. I just haven’t had the options to use it without making some modifications.

Moving away from chai would require significantly more effort, and I try to avoid taking steps like that. While placing it in the right place is something that can be done, I haven't bothered to fork the chai yet.

It's completely fine if you choose not to merge my changes.

However, I need a trustworthy user script runner. Would you mind if I fork it and publish it in the Mozilla Store?  

Thank you for bringing my mistakes to my attention.

arantius:
> The chai software is licensed under the MIT License, and ...

What I'm saying is we have an established mechanism for vendoring third party dependencies.  Any open source contribution should be made to follow existing patterns.

Take a look at e.g.:

https://github.com/greasemonkey/greasemonkey/tree/master/third-party/tinybind

A `README` explains exactly what it is, where it came from, and the license terms.  There should at a minimum be one version checked in identical to that external version, then if changes are made they should be in a separate commit.  (Though we don't have any existing examples of changes like this.)

Soldy:
In the long term, I am considering replacing Karma with Puppeteer since Karma has been [deprecated](https://github.com/karma-runner/karma/blob/master/README.md). This should resolve the current issues, but it will require about 3-4 hours of work or more. My previous solution was a temporary fix. My main priorities are to [update the code mirror, implement a dark mode, and support Vim and Emacs key layouts](https://github.com/Soldy/greasemonkey/tree/codemirror). However, without a test, this has a significant risk. I have also identified some [minor problems](https://github.com/Soldy/greasemonkey/tree/asyncrun), and I have started working on a cleaner (though not yet fully developed) [meta parser](https://github.com/Soldy/greasemonkey/tree/newparse). 
I suggest creating an additional directory, similar to a "third-party" folder for testing purposes, or a [fork of the chai](https://github.com/Soldy/Chai-for-karma) I did and you have [done as well with sinon](https://github.com/arantius/karma-sinon-chrome). I believe that Chai.js should not be included in the distributed package, so it cannot be in the third-party directory.

PR #3220, opened 2025-08-30
Fix GM.xmlHttpRequest when privacy.firstparty.isolate is true

### LINKED ISSUE #2985: privacy.firstparty.isolate=true breaks GM.xmlHttpRequest
I get **cookies is undefined** and **Unchecked lastError value: Error: First-Party Isolation is enabled, but the required 'firstPartyDomain' attribute was not set** when running a GM userscript calling `GM.xmlHttpRequest()` with `privacy.firstparty.isolate` ([First-Party Isolation](https://www.bleepingcomputer.com/news/software/another-tor-browser-feature-makes-it-into-firefox-first-party-isolation/)) enabled (set to `true`):

gm xmlhttprequest privacy firstparty isolate

For now, this means I'm disabling First-Party Isolation, but I'm hoping there's a solution or workaround that will make this work better.

In the event it matters (it shouldn't, I'm on FF 61), I have followed the [ghacks workaround](https://github.com/ghacksuserjs/ghacks-user.js/wiki/4.1.1-Setting-Extension-Permission-Exceptions) to explicitly allow GM to set cookies.

Scripter17:
Fixes #2985

### DIFF
--- a/src/bg/on-user-script-xhr.js
+++ b/src/bg/on-user-script-xhr.js
@@ -125,7 +125,7 @@ function open(xhr, d, port, tabUrl) {
 
   // If this is a same-origin XHR or the user opted in with withCredentials,
   // add cookies unless already specified by the user.
-  chrome.cookies.getAll({url: d.url}, cookies => {
+  chrome.cookies.getAll({url: d.url, firstPartyDomain: tabUrl.hostname}, cookies => {
     if (cookies.length && !hasCookieHeader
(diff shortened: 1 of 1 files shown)


woodsamy544 (approved):

VladPetrariu (changes requested):
implement the registrable-domain handling and add a compatibility fallback and tests, then I’d approve.

Scripter17:
Somehow only just saw your comment. How would I do those?

PR #3231, opened 2026-05-22
Added @include with url of active tab

Cerberus-tm:
Gets the url of the active tab and adds @include with the url to the template for new scripts. I am new to editing/committing stuff on Github and webextensions, so I hope this works. Better check it just in case.

### DIFF
--- a/src/browser/monkey-menu.js
+++ b/src/browser/monkey-menu.js
@@ -411,17 +411,18 @@ function navigateToScript(uuid) {
 function newUserScript() {
   let r = Math.floor(Math.random() * 900000 + 100000);
   let name = _('unnamed_script_RAND', r);
-  let include = '';
+  let match_url = '*';
   let active_tabs = chrome.tabs
       .query({'active': true, 'currentWindow': true});
-  if (active_tabs && active_tabs[0] && active_tabs[0].url) {
-      include = `// @include  ${active_tabs[0].url}\n`;
+  if (active_tabs[0]?.url) {
+      match_url = active_tabs[0].url;
   }
-  let scriptSource = `// ==UserScript==
+  let scriptSource =
+`// ==UserScript==
 // @name     ${name}
 // @version  1
-` + include +
-`// @grant    none
+// @match    ${match_url}
+// @grant    none
 // ==/UserScript==`;
   let downloader
       = new UserScriptDownloader().setScriptContent(scriptSource);
--- a/src/browser/monkey-menu.js
+++ b/src/browser/monkey-menu.js
@@ -411,12 +411,9 @@ function navigateToScript(uuid) {
 function newUserScript() {
   let r = Math.floor(Math.random() * 900000 + 100000);
   let name = _('unnamed_script_RAND', r);
-  let match_url = '*';
   let active_tabs = chrome.tabs
       .query({'active': true, 'currentWindow': true});
-  if (active_tabs[0]?.url) {
-      match_url = active_tabs[0].url;
-  }
+  let match_url = active_tabs[0]?.url || '<all_urls>';
   let scriptSource =
 `// ==UserScript==
 // @name     ${name}
--- a/src/browser/monkey-menu.js
+++ b/src/browser/monkey-menu.js
@@ -411,10 +411,17 @@ function navigateToScript(uuid) {
 function newUserScript() {
   let r = Math.floor(Math.random() * 900000 + 100000);
   let name = _('unnamed_script_RAND', r);
+  let include = '';
+  let active_tabs = chrome.tabs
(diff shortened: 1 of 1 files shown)


src/browser/monkey-menu.js
```diff
@@ -411,10 +411,17 @@ function navigateToScript(uuid) {
 function newUserScript() {
   let r = Math.floor(Math.random() * 900000 + 100000);
   let name = _('unnamed_script_RAND', r);
+  let include = '';
+  let active_tabs = chrome.tabs
+      .query({'active': true, 'currentWindow': true});
+  if (active_tabs && active_tabs[0] && active_tabs[0].url) {
```
arantius: I think I'd write this simply `if (active_tabs[0]?.url)`

src/browser/monkey-menu.js
```diff
@@ -411,10 +411,17 @@ function navigateToScript(uuid) {
...
+      include = `// @include  ${active_tabs[0].url}\n`;
```
arantius: I think it's probably more common to write a script for a whole site than for just one particular URL?

arantius: When compatible I'd suggest `@match` over `@include` -- and here I think it's compatible.

src/browser/monkey-menu.js
```diff
@@ -411,10 +411,17 @@ function navigateToScript(uuid) {
...
+  }
   let scriptSource = `// ==UserScript==
 // @name     ${name}
 // @version  1
-// @grant    none
+` + include +
```
arantius: This is already in a backtick string, so just `${include}` ?  I suppose that has the chance to leave an empty line?  Default to an explicit `*` if we can't be more specific?

src/browser/monkey-menu.js
Cerberus-tm: Yeah, I wanted to do that, but I was slightly afraid I might have misunderstood something about what happens in JS if you try to access an item in an array that doesn't exist, was afraid of some blocking error so played it extra safe. But then let's do ` if (active_tabs[0]?.url)`.

src/browser/monkey-menu.js
Cerberus-tm: That is true, I thought about that as well; but it is not so rare to want more than just the domain, and deleting the end is much easier for the user than adding it back, was my thought.

Cerberus-tm: OK I wasn't sure how `@match` behaved, let's do that.

src/browser/monkey-menu.js
Cerberus-tm: Right, I did it this way because otherwise we'd have to do either a line `// @version 1${include}` or a line `${include}// @grant none`, which seemed a bit clunky.

Explict * sounds like a good idea, I will do that. Let's see if I can figure out how to change the commit request in Github.

src/browser/monkey-menu.js
```diff
@@ -411,9 +411,17 @@ function navigateToScript(uuid) {
 function newUserScript() {
   let r = Math.floor(Math.random() * 900000 + 100000);
   let name = _('unnamed_script_RAND', r);
-  let scriptSource = `// ==UserScript==
+  let match_url = '*';
```
arantius: Oh, phooey.

* https://wiki.greasespot.net/Metadata_Block#@match
* https://developer.chrome.com/docs/extensions/develop/concepts/match-patterns

`@match *` is not valid.  For match it's `<all_urls>`.  (Match syntax is more complicated and I didn't remember it myself.

src/browser/monkey-menu.js
```diff
@@ -411,9 +411,17 @@ function navigateToScript(uuid) {
...
+  let active_tabs = chrome.tabs
+      .query({'active': true, 'currentWindow': true});
+  if (active_tabs[0]?.url) {
```
arantius: Accidentally said this in the commit, but a much shorter/more idiomatic form would be:

`let match_url = active_tabs[0]?.url || '<all_urls>';`

src/browser/monkey-menu.js
Cerberus-tm: I should have checked that. Will switch to `<all_urls>`.

src/browser/monkey-menu.js
Cerberus-tm: I considered it but wasn't 1000% certain how that behaved so I was probably being too safe. Will use that.

arantius:
Thanks for contributing!

Cerberus-tm:
Thanks for helping me!

I edited the file in my fork of the repository each time, then pressed the buttons that appeared, to push it to the master branch. I hope the result was the proper way / wasn't inconvenient for others; I don't know how Github works.

